PR #1, opened 2022-12-14
Add support to emit links for needed types

ianoc:
Start looking at what a type inherits from, the return types of its methods + args. And emit links between the class itself and these types.

### DIFF
--- a/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
+++ b/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
@@ -516,11 +516,16 @@ object ScalaSourceEntityExtractor {
       val stringRefs = refs.map(_.toString)
       val links = getLinksRecursive(s, Nil).iterator
         .flatMap { case (src, dests) =>
-          dests.iterator
-            .map(_.toString)
-            .iterator
+          dests.iterator.iterator
+            .map(_.value)
             .flatMap { e =>
-              stringRefs.iterator.filter { y => y.endsWith(e) }
+              val data = e.split('.').toList
+              val h :: tail = data
+              stringRefs.iterator.filter { y => y.endsWith(h) }.flatMap { e =>
+                e :: tail.map { t =>
+                  s"${e}.$t"
+                }
+              }
             }
             .map { linkTarget =>
               s"link: ${src} -> $linkTarget"
@@ -557,7 +562,11 @@ object ScalaSourceEntityExtractor {
       case Term.Param(mods, name, optType, optTerm) =>
         log(s"treeToNames: Term.Param($mods, $name, $optType, $optTerm)")
         optType.toList.map(typeToNames).flatten
+      case Init(tpe, name, terms) =>
+        log(s"Init($tpe, $name, $terms)")
+        typeToNames(tpe)
       case _ => Nil
+
       // case t =>
       // sys.error(s"unknown: ${t.getClass} $t")
     }
@@ -570,10 +579,11 @@ object ScalaSourceEntityExtractor {
       case Type.Apply(left, args) =>
         log(s"typeToNames: Type.Apply($left, $args)")
         (left :: args).traverse(typeToNames).flatten
-      // sys.error(s"unknown: ${tpe.getClass} $tpe")
+      case s @ Type.Select(_, _) =>
+        List(Name(s.toString))
       case _ => Nil
       // case t =>
-      //   sys.error(s"unknown: ${t.getClass} $t")
+      // sys.error(s"unknown: ${t.getClass} $t")
     }
   }
 
@@ -837,9 +847,12 @@ object ScalaSourceEntityExtractor {
         log(s"Name($n)")
         if (n.nonEmpty) referTo(nm)
         else unitEnv
-      case Pkg.Object(mods, name, template) =>
+      case Pkg.Object(mods, name, template @ Template(_, init, _, stats)) =>
         log(s"package object: mods = $mods, name = $name, template = $template")
-        define(name) *> inside(name, inspect(template))
+        link(
+          name,
+          init.map(_.tpe).flatMap(typeToNames) ++ stats.flatMap(treeToNames)
+        ) *> inside(name, inspect(template))
       case Template(early, init, self, stats) =>
         log(
           s"Template(early = $early, init = $init, self = $self, stats = $stats)"
@@ -865,12 +878,12 @@ object ScalaSourceEntityExtractor {
               .traverse_(inspect) *> stats.traverse_(inspect(_))
           }
         )
-      case Defn.Object(mods, name, templ @ Template(_, init, _, _)) =>
+      case Defn.Object(mods, name, templ @ Template(_, init, _, stats)) =>
         log(s"Object($mods, $name, $templ)")
 
         define(name) *> link(
           name,
-          init.map(_.tpe).flatMap(typeToNames)
+          init.flatMap(treeToNames) ++ stats.flatMap(treeToNames)
         ) *> inside(name, inspect(templ))
       case Defn.Val(mods, pats, optType, term) =>
         log(s"Val($mods, $pats, $optType, $term)")
@@ -905,10 +918,13 @@ object ScalaSourceEntityExtractor {
             name,
             tparams,
             ctor,
-            templ @ Template(_, init, _, _)
+            templ @ Template(_, init, _, stats)
           ) =>
         log(s"Defn.Trait($mods, $name, $tparams, $ctor, $templ)")
-        define(name) *> link(name, init.map(_.tpe).flatMap(typeToNames)) *>
+        define(name) *> link(
+          name,
+          init.map(_.tpe).flatMap(typeToNames) ++ stats.flatMap(treeToNames)
+        ) *>
           inside(
             name, {
               tparams.traverse_ { t =>
--- a/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
+++ b/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
@@ -181,6 +181,7 @@ object ScalaSourceEntityExtractor {
           .to(SortedSet)
 
       val directRefs = resolveNonLocals(refs)(_.value)
+
       val imps = resolveNonLocals(
         allUnresolvedImportEntities
           .flatMap(_.prefixes.toList)
@@ -267,16 +268,22 @@ object ScalaSourceEntityExtractor {
   def define(n: Name): Env[Unit] =
     updateState(_.addDef(n))
 
-  def link(n: Name, target: List[Name]): Env[Unit] = {
-    if (target.nonEmpty) {
-      updateState(_.addLink(n, NonEmptyList.fromListUnsafe(target)))
+  def link(n: Name, targets: List[List[Name]]): Env[Unit] = {
+    if (targets.nonEmpty) {
+      targets.traverse_ { t =>
+        if (t.nonEmpty) {
+          updateState(_.addLink(n, NonEmptyList.fromListUnsafe(t)))
+        } else {
+          updateState(identity)
+        }
+      }
     } else {
       State.empty
     }
   }
 
   def link(n: Name, target: Name): Env[Unit] =
-    link(n, target :: Nil)
+    link(n, List(target :: Nil))
 
   // val Foo(bar, baz) = ...
   // so we need to define all the names we hit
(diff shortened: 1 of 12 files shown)


johnynek:
I don't see that we are running the JVM tests in CI. Are we?

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -247,6 +267,17 @@ object ScalaSourceEntityExtractor {
   def define(n: Name): Env[Unit] =
     updateState(_.addDef(n))
 
+  def link(n: Name, target: List[Name]): Env[Unit] = {
+    if (target.nonEmpty) {
```
johnynek, line 271: nit:
```scala
NonEmptyList.fromList(target) match {
  case Some(nel) => updateState(_.addLink(n, nel))
  case None => State.empty
}
```

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -470,8 +513,28 @@ object ScalaSourceEntityExtractor {
     def processScope(s: ScopeState): Writer[DataBlock, Unit] = {
       val defs = s.definedEntities
       val refs = s.nonLocalRefs(getResolve(s))
-
-      Writer.tell(DataBlock("", defs = defs, refs = refs))
+      val stringRefs = refs.map(_.toString)
+      val links = getLinksRecursive(s, Nil).iterator
+        .flatMap { case (src, dests) =>
+          dests.iterator.iterator
```
johnynek, line 519: I am pretty sure we don't want `.iterator.iterator`. I think `dests` is a `List` so `List.iterator` is an iterator.

also, I think a for comprehension will actually read better here:

```scala
val linksIter = for {
  sd <- getLinksRecursive(s, Nil).iterator
  (src, dests) = sd
  dest <- dests.iterator
  data = dest.value.split('.').toList
  y <- stringRefs.iterator
  if y.endsWith(data.head)
  tailStr = tail.iterator.map { t => s"${y}.$t" }
  linkTarget <- Iterator.single(e) ++ tailStr
} yield s"link: ${src} -> $linkTarget"

val links = linksIter.to(SortedSet)
```

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+  }
+
+  def treeToNames(tpe: scala.meta.Tree): List[Name] = {
+    import scala.meta._
+    tpe match {
+      case Defn.Def(mods, name, tparams, paramss, optType, body) =>
+        log(
+          s"treeToNames:  Defn.Def(mods:$mods, name: $name, tparams: $tparams, paramss: $paramss, optType: $optType, body: $body)"
+        )
+        optType.toList.map(typeToNames).flatten ++
```
johnynek, line 560: nit: `map(fn).flatten == flatMap(fn)`

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+          paramss.flatMap(_.map(treeToNames)).flatten
```
johnynek, line 561: nit: `paramss.flatMap(_.map(treeToNames)).flatten == paramss.flatMap(_.flatMap(treeToNames))`

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+      case Term.Param(mods, name, optType, optTerm) =>
+        log(s"treeToNames: Term.Param($mods, $name, $optType, $optTerm)")
+        optType.toList.map(typeToNames).flatten
```
johnynek, line 564: map.flatten => flatMap

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+      case Init(tpe, name, terms) =>
+        log(s"Init($tpe, $name, $terms)")
+        typeToNames(tpe)
+
+      case Defn.Val(mods, pats, optType, term) =>
```
johnynek, line 569: I think it is possible the `pats` could include types as: `val (foo: Int, bar: String) = tuple` I think.... it may be rare, but I think that would be ignored currently.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+      case _ => Nil
+
+      // case t =>
+      // sys.error(s"unknown: ${t.getClass} $t")
+    }
+  }
+  def typeToNames(tpe: scala.meta.Type): List[Name] = {
+    import scala.meta._
+    tpe match {
+      case tn @ Type.Name(_) =>
```
johnynek, line 583: nit: `tn :: Nil` is possibly faster since it is fewer allocations. `List(tn)` does `List.apply(Seq(tn))` basically.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -482,17 +545,61 @@ object ScalaSourceEntityExtractor {
...
+        List(tn)
+      case Type.Apply(left, args) =>
+        log(s"typeToNames: Type.Apply($left, $args)")
+        (left :: args).traverse(typeToNames).flatten
+      case s @ Type.Select(_, _) =>
+        List(Name(s.toString))
```
johnynek, line 589: `Name(s.toString) :: Nil` is maybe better, but I don't quit get this. This is `Foo.Bar` right? I don't quite remember if we used `Entity` here to break the dots up...

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -746,9 +853,12 @@ object ScalaSourceEntityExtractor {
         log(s"Name($n)")
         if (n.nonEmpty) referTo(nm)
         else unitEnv
-      case Pkg.Object(mods, name, template) =>
+      case Pkg.Object(mods, name, template @ Template(_, init, _, stats)) =>
         log(s"package object: mods = $mods, name = $name, template = $template")
-        define(name) *> inside(name, inspect(template))
```
johnynek, line 751: I see `define(name)` is removed. I wonder if any `Object` needs to define that name, or can we always get away with just the members?

I guess we do, since you could have:

```scala
object EmptyIt extends Iterator[Nothing] {
...
}

callSomeIteratorFn(EmptyIt)
```
but that pattern is basically never used for package objects I guess... but I suppose it could be?

Anyway, we can add tests for the thing we need later if it causes problems.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -761,16 +871,26 @@ object ScalaSourceEntityExtractor {
...
             (tparams ::: ctor :: template :: early ::: init ::: self :: Nil)
               .traverse_(inspect) *> stats.traverse_(inspect(_))
           }
         )
-      case Defn.Object(mods, name, templ) =>
+      case Defn.Object(mods, name, templ @ Template(_, init, _, stats)) =>
         log(s"Object($mods, $name, $templ)")
-        define(name) *> inside(name, inspect(templ))
+
+        define(name) *> link(
```
johnynek, line 890: so the only difference here is define(name) or not, right? Could we add a comment to remind us later?

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DataApi.scala
```diff
@@ -23,6 +23,7 @@ final case class DataBlock(
 ) {
   def addDef(e: Entity) = copy(defs = defs + e)
   def addRef(e: Entity) = copy(refs = refs + e)
+  def addLink(src: Entity, dests: List[Entity]) = this
```
johnynek, line 26: I don't get this. It seems like we did all this work to ignore the links?

language_generators/scala-defref-extractor/src/test/scala/io/bazeltools/buildgen/scaladefref/CanParseFileTest.scala
```diff
@@ -17,41 +17,62 @@ class CanParseFileTest extends AnyFunSuite {
...
+        Entity.dotted("com"),
+        Entity.dotted("com.example"),
+        Entity.dotted("com.example.Elephant"),
+        Entity.dotted("com.example.Wolf"),
+        Entity.dotted("com.foo.bar.com"),
+        Entity.dotted("com.foo.bar.com.example"),
+        Entity.dotted("com.foo.bar.com.example.Elephant"),
+        Entity.dotted("com.foo.bar.com.example.Wolf")
+      ),
+      SortedSet(
```
johnynek, line 69: I'm not following how this is populated now since DataBlock ignores addLink.

Oh... is this test actually run in CI? I don't think so.

ianoc:
> I don't see that we are running the JVM tests in CI. Are we?

Nope, thats not wired in

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
ianoc: Yep we need a test around this and some more concrete argument noted in the test. And thats right we could have it with package objects but it feels rarer. Vs the other side where it causes huge links

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
ianoc: This is wrong yeah :/

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -248,6 +269,23 @@ object ScalaSourceEntityExtractor {
   def define(n: Name): Env[Unit] =
     updateState(_.addDef(n))
 
+  def link(n: Name, targets: List[List[Name]]): Env[Unit] = {
+    if (targets.nonEmpty) {
```
johnynek, line 273: I don't think you need the `targets.nonEmpty` since traversing an empty list gives you just the empty `Env[Unit]`

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -290,6 +328,31 @@ object ScalaSourceEntityExtractor {
...
+        case n @ Term.Name(_)                    => n :: acc
+        case Term.Select(left, n @ Term.Name(_)) => loop(left, n :: acc)
+        case Term.Super(thisp, superp) =>
+          log(s"Term.Super(thisp: $thisp, superp: $superp)")
+          // This should be safe, since we should have a link onto our parent type already. And with a super reference we would need to resolve this
+          // onto figuring out our type/things we are inheriting from?
+          Nil
+        case _ =>
+          sys.error(
+            s"Un expected term : $t  (class: ${t.getClass.getName} )hit when trying to unroll outer term: $outerTerm"
```
johnynek, line 344: can we fix the weird spacing? `Un expected => Unexpected` and `)hit => ) hit`.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -290,6 +328,31 @@ object ScalaSourceEntityExtractor {
...
+          )
+      }
+    }
+
+    NonEmptyList.fromListUnsafe(loop(outerTerm, Nil) ::: n :: Nil)
```
johnynek, line 349: rather than `fromListUnsafe` can we use: `NonEmptyList(loop(outerTerm, Nil), n :: Nil)`

ianoc:
this hasn't really solved this problem alas, so going to abandon now. Seems like the right way to solve in bazel ++ rules_scala

PR #5, opened 2022-12-14
We should include the path in the sha, since its inside the produced …

ianoc:
…json

### DIFF
--- a/crates/driver/src/extract_defrefs.rs
+++ b/crates/driver/src/extract_defrefs.rs
@@ -49,6 +49,7 @@ pub struct Extractors(pub HashMap<String, Extractor>);
 #[derive(Debug, Clone)]
 pub struct Extractor {
     pub path: PathBuf,
+    pub extractor_sha: Sha256Value
 }
 
 #[derive(Debug)]
@@ -76,11 +77,15 @@ async fn process_file(
             )
         })?;
         use std::os::unix::ffi::OsStrExt;
-        // The outputs include the entity path/label information of the target
+        // The input of the relative path is carried through into the output result
         // so we need to include this in our sha we use to identify the file.
+        //
+        // We also include the sha of the extractor itself
+        //
         Sha256Value::hash_iter_bytes(
             vec![
                 r.as_bytes(),
+                opt.extractor.extractor_sha.as_bytes(),
                 relative_path.as_os_str().as_bytes()
             ].into_iter())
     };
@@ -402,7 +407,7 @@ async fn run_extractors_on_data<'a>(
     Ok((results, (max_target, max_duration)))
 }
 
-fn load_extractors(extract: &'static Extract) -> Result<Extractors> {
+async fn load_extractors(extract: &'static Extract) -> Result<Extractors> {
     let mut r = HashMap::default();
     for combo in extract.extractor.iter() {
         let p: Vec<&str> = combo.split(':').collect();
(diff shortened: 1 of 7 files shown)


johnynek (approved):

crates/driver/src/extract_defrefs.rs
```diff
@@ -75,7 +75,14 @@ async fn process_file(
                 e
             )
         })?;
-        r
+        use std::os::unix::ffi::OsStrExt;
+        // The outputs include the entity path/label information of the target
```
johnynek, line 79: isn't it the *inputs* that matter that you are passing to the command below.

crates/driver/src/extract_defrefs.rs
```diff
@@ -49,6 +49,7 @@ pub struct Extractors(pub HashMap<String, Extractor>);
 #[derive(Debug, Clone)]
 pub struct Extractor {
     pub path: PathBuf,
+    pub extractor_sha: Sha256Value
```
johnynek, line 52: 💯

crates/driver/src/extract_defrefs.rs
```diff
@@ -105,7 +110,7 @@ async fn process_file(
         command
             .arg("--working-directory")
             .arg(working_directory.as_path());
-        command.arg("--label-or-repo-path").arg(path.as_path());
+        command.arg("--label-or-repo-path").arg(relative_path.as_path());
```
johnynek, line 113: seems weird that both the `--relative-input-paths` and `--label-or-repo-path` are the same values. Can we add a comment as to why we can't have a single argument?

nit: why are we sometimes doing
```rust
command
  .arg("...")
  .arg("...")
```
and sometimes: `command.arg("--key").arg("value")`

I guess I slightly prefer the second, but would probably just say we should have one style.

crates/driver/src/extract_defrefs.rs
```diff
@@ -447,7 +453,7 @@ pub async fn extract_defrefs(
     let sha_of_conf: Sha256Value = merged_config_str.as_bytes().into();
     let sha_of_conf_config = Arc::new(format!("{}", sha_of_conf));
 
-    let extractors = load_extractors(extract)?;
+    let extractors = load_extractors(extract).await?;
 
     let sha_to_extract_root = Box::leak(Box::new(opt.cache_path.join("sha_to_extract")));
```
johnynek, line 458: leak sounds scary... I better look that up...

johnynek (approved):
a couple of minor nits.

Good catch including the sha of the executable (although that does seem to imply we really need fully statically linked binaries for this to be truly safe).

crates/driver/src/extract_defrefs.rs
ianoc: oh. am . i guess its just a formatting thing to do with line length picking between them? I've not put any thought into it.

The label-or-repo-path doesn't have a file path always associated with it, it can also be a label (for external dependencies).

crates/driver/src/extract_defrefs.rs
ianoc: Leak is fun, its not `unsafe` since it tells rust to take an allocated on the heap thing and just forget about cleaning it up. So then we can guarantee (without doing something unsafe), that it will survive for the length of the application. So we can get a static reference to it. Its quite handy for config/command line apps, rather than having to use arc's/ or caring about reference lifetimes we can just pass it around as an immutable static value

ianoc:
Yeah thats true, static binaries is maybe a stronger guarantee(bazel doesn't have this one). but yes if its an entry point that follows on and shells out to different app/or it points into a folder somehow then yep this sha cannot capture that :s

PR #7, opened 2022-12-16
Add most of the tests from original code

johnynek:


language_generators/scala-defref-extractor/src/test/scala/io/bazeltools/buildgen/javadefref/JavaSourceEntityExtractorTest.scala
```diff
@@ -0,0 +1,114 @@
...
+    val expected = DefsRefs(
+        defs = ents("""["foo.bar.Bar"]"""),
+        // TODO: if we call b.call inside a method we shouldn't see b as a dep
+        //refs = ents("""["b", "foo.bar.b", "foo.bar.Baz", "Baz", "foo.bar.Pop", "Pop", "Bippy", "foo.bar.Bippy", "some.Pack"]"""))
+        refs = ents("""["foo.bar.Baz", "Baz", "foo.bar.Pop", "Pop", "Bippy", "foo.bar.Bippy", "some.Pack"]"""))
+
+    got.assertEquals(expected)
+  }
+
+  /*
```
johnynek, line 48: this test currently fails. It seems unref is not working in the new code, so we may not be able to remove phantom references.

johnynek:
there are a number of scala parsing tests to port over also, but those are a bit more work to port over because they are mixing in parts of the code that didn't make it into the rust rewrite.

Still I think we should probably do it because those examples were some hard ones to get working correctly and possibly easy to suffer regressions on.

language_generators/scala-defref-extractor/src/test/scala/io/bazeltools/buildgen/javadefref/JavaSourceEntityExtractorTest.scala
ianoc: I mentioned this in slack, but the semantic knowledge of what ref/unref and so on doesn't exist in the extractors at all anymore. I didn't include unref i don't think since we haven't been using it. But that handling occurs downstream

ianoc (approved):
I think we can either remove the commented test, or change that it looks for to be the extracted directives.

language_generators/scala-defref-extractor/src/test/scala/io/bazeltools/buildgen/javadefref/JavaSourceEntityExtractorTest.scala
johnynek: okay, I fixed the test and will merge when it goes green.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/PathTree.scala
```diff
@@ -76,60 +92,65 @@ object PathTree {
         case head :: tail =>
           branches.get(head) match {
             case Some(n) => n.getAt(tail)
-            case None    => None
+            case None => None
```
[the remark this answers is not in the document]
johnynek: I didn't run it because I was afraid it would rewrite everything, but I should have just tried after a commit. I see that you have run it already so it doesn't rewrite the repo.

We should probably add a CI test to require it.

PR #12, opened 2023-01-05
Test scalafmt in CI

johnynek:


language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DriverApplication.scala
```diff
@@ -28,7 +28,7 @@ abstract class DriverApplication extends IOApp {
       )
     }
 
-  private[this] def extractDataBlocks(
+     private[this] def extractDataBlocks(
```
johnynek, line 31: this should fail the CI. If it does, I'll fix and then we can merge.

ianoc (approved):

PR #13, opened 2023-02-14
Protobuf support

eed3si9n:
This uses tree-sitter-proto to parse the Procol Buffer schema file for Protobuf extractor.

In addition, since `proto_library` typically requires accompanying
rules, I've added `secondary_rules` map, which will generate
rules within the same `BUILD.bazel` (as opposed to relying on macros).
The dependency relationship between the main target and the secondaries
can be specified using `extra_key_to_list` as

```json
"deps": [":${name}"]
```

I've also refactored the BUILD generation function so we can actually write a unit test, AND I've also added an example Bazel repository.


johnynek:
This is exciting! Can you add a comment if we keep the sub module approach. I guess the crate isn’t published yet or something?

eed3si9n:
> This is exciting! Can you add a comment if we keep the sub module approach. I guess the crate isn’t published yet or something?

Yea. tree-sitter culture is everyone grabs `*.c` code for the parser you want and bind it into Vim etc, so crates are often not really maintained.

ianoc:
Can we not use cargo's support to refer to git dependencies rather than mixing in the submodules?

ianoc:
```
tree-sitter-proto = { git = "https://github.com/mitchellh/tree-sitter-proto.git", rev = "42d82fa18f8afe59b5fc0b16c207ee4f84cb185f"}
```

I checked out a copy of your work to confirm this works -- but you can use this for the version specification. We can't publish to crates.io if we have git deps -- but we already do for the rust python support. So i've no problem adding more

ianoc:
I transferred the repo now @eed3si9n so you should able to have full control/develop with a branch if you like too

crates/protobuf_extractor/src/extract_protobuf_imports.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#[derive(Debug, PartialEq)]
+pub struct ProtobufSource {
+    pub imports: Vec<String>
+}
+
+pub fn parse(source: &str, _source_path: &str) -> Result<ProtobufSource> {
+    let mut buf = Vec::default();
+    let mut parser = Parser::new();
+    parser.set_language(tree_sitter_proto::language()).expect("tree-sitter failed to initialize");
```
ianoc, line 12: Should avoid the expect and can return the Err branch on the anyhow

crates/protobuf_extractor/src/extract_protobuf_imports.rs
```diff
@@ -0,0 +1,56 @@
...
+    let tree = parser.parse(source, None).expect("parse failed");
+    let root_node = tree.root_node();
+    let mut cursor = root_node.walk();
+    let bytes = source.as_bytes();
+    for child_node in root_node.children(&mut cursor) {
+        match child_node.kind() {
+            "import" =>
+                {
+                    for path in child_node.children_by_field_name("path", &mut child_node.walk()) {
+                        let mut quoted = path.utf8_text(bytes).unwrap().chars();
```
ianoc, line 22: avoiding the unwrap too as above

crates/protobuf_extractor/src/extract_protobuf_imports.rs
```diff
@@ -0,0 +1,56 @@
...
+                        // Remove the quotation marks
```
ianoc, line 23: This is interesting in the level of parsing vs not in this. -- the lack of quotes extraction. Could you try extend this next maybe to try get a set of the messages/services defined in the file? curious how easy/hard or possibly error prone or not that is. (i guess how much is left to the app here vs the tree sitter handles for us i'm curious to see)

johnynek: if we are only parsing the imports, I wonder if we are winning that much by using tree-sitter (which *seems* to me pretty low level still).

On the other hand, we probably eventually want all the message names and the package as well very soon, so doing all that with nom may be a pain. But seems like maybe close (and we would also have an easier time adding features for linting, etc...)

eed3si9n: tree-sitter on its own is mostly a parser generator with some runtime. The actual grammar comes from mitchellh/tree-sitter-proto, which defined the `import` tree as https://github.com/mitchellh/tree-sitter-proto/blob/42d82fa18f8afe59b5fc0b16c207ee4f84cb185f/grammar.js#L34-L39

```js
    import: $ => seq(
      'import',
      optional(choice('weak', 'public')),
      field('path', $.string),
      ';',
    ),
```

In my code `path.utf8_text(bytes)` basically extracts the substring of what's marked as `'path'`, which in this case is a `$.string` node, which is completely unaware of quotation because each language does slightly different things with it (single quote, triple quote, string interpolation etc).

eed3si9n: The benefit of tree-sitter-based parsers is that the run-time is usually much faster than hand parsing with regex because the state machine has been expanded out in C code, also it supports partial parsing and ignore the parts that it doesn't understand. Eventually I think this could be used for Scala parsing as well.

We also have to parse comments if we want to support directives.

crates/protobuf_extractor/src/extract_protobuf_imports.rs
eed3si9n: Yesterday I learned `?` - c48f242a73f03414855a3551ca71d70ebbd6e722

weichunstevelee:
what's the latest on this PR? should we try to merge this PR? As we can have more and more protobuf pretty soon, I'm looking to bridge the final gap. Happy to help here too

eed3si9n:
> what's the latest on this PR? should we try to merge this PR? As we can have more and more protobuf pretty soon, I'm looking to bridge the final gap. Happy to help here too

I haven't looked at it in a while, but we can resurrect it.

DEVELOPING.md
eed3si9n, line 1: I mostly moved this from `crates/README.md` with some updates on the details.

ianoc:
Once this is happy and moving somewhat could we split it up amongst the readme moving/refactoring and testing changes vs things proto itself needs ?

eed3si9n:
> Once this is happy and moving somewhat could we split it up amongst the readme moving/refactoring and testing changes vs things proto itself needs ?

Yea, I guess so. More than half the changes are test related. But given the actually working `example/` monorepo, it feels like an open source project since we can show how it's supposed to be all hooked up, and review with more confidence. The CI is green, so if everyone is somewhat happy with the direction of this PR I can split this into smaller PRs or commits.

eed3si9n:
I've split out example repo to https://github.com/bazeltools/bzl-gen-build/pull/89

example/src/main/protos/com/example/foo/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+load('@rules_proto//proto:defs.bzl', 'proto_library')
```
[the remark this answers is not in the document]
eed3si9n: At least right now, I don't think there's a feature to support generating multiple main targets per source file. I am guessing the workaround is having one proto file per directory?

crates/driver/src/print_build.rs
```diff
@@ -542,14 +548,9 @@ async fn print_file(
...
 
-                let _handle = concurrent_io_operations.acquire().await?;
                 let sub_target = opt.working_directory.join(node).join("BUILD.bazel");
                 emitted_files.push(sub_target.clone());
-                tokio::fs::write(&sub_target, t.emit_build_file()?)
-                    .await
-                    .with_context(|| {
-                        format!("Attempting to write file data to {:?}", target_file)
-                    })?;
+                on_child(sub_target, t);
```
ianoc, line 553: Does this do anything? `on_child` returns a future that we discard, so it would be a no-op i think?

eed3si9n: Good catch. on_child is the callback, so it should be `on_child(sub_target, t)?;`?

crates/driver/src/print_build.rs
```diff
@@ -474,6 +474,10 @@ where
     }
 
     apply_attr_string_lists(&mut extra_kv_pairs, &graph_node.node_metadata);
+    let deps = extra_kv_pairs
```
johnynek, line 477: can we comment why we are taking this here? Is it because extra_kv_pairs is being mutated below?

I am having a hard time following this function since it is very large IIUC and has a lot of nested internal functions.

crates/driver/src/print_build.rs
```diff
@@ -571,6 +575,89 @@ where
...
         t.entries.push(target);
     }
+
+    fn apply_secondary_rules(
+        target_entries: &mut TargetEntries,
+        module_config: &ModuleConfig,
+        parent_target_name: &str,
+        parent_include_src: &Vec<String>,
+        parent_deps: &Vec<String>,
+    ) -> Result<()> {
```
johnynek, line 585: I'm not seeing how this method can ever fail. Why does it return anything but `()`?

crates/protobuf_extractor/src/extract_protobuf_imports.rs
```diff
@@ -0,0 +1,62 @@
...
+        };
+        let root_node = tree.root_node();
+        let mut cursor = root_node.walk();
+        let bytes = source.as_bytes();
+        for child_node in root_node.children(&mut cursor) {
+            match child_node.kind() {
+                "import" =>
+                    {
+                        for path in child_node.children_by_field_name("path", &mut child_node.walk()) {
+                            let mut quoted = path.utf8_text(bytes).unwrap().chars();
```
johnynek, line 26: this unwrap is a panic if it fails right? Can we instead add it into the `Result` type we are returning? (maybe just matching and `bail!` as above?)

crates/protobuf_extractor/src/main.rs
```diff
@@ -0,0 +1,110 @@
...
+                    "Error while parsing file {:?
+        }",
+                    input_file
+                )
+            })?;
+
+            refs.extend(program.imports);
+        }
+
+        defs.extend(Some(relative_path.clone()));
```
johnynek, line 90: if I understand, here we are using the relative file paths as the `refs` and `defs` is that right? Can we add a comment in that this is different from the other cases where we are using symbol names as the things we are ref or defing.

johnynek (approved):
I had a couple of nits and requests for comments, but I think I understand what's going on.

If you could add a few requested comments I'd appreciate it.

crates/driver/src/print_build.rs
eed3si9n: Addressed in c97d4b7c8d900f53d63d9e54182d7ffbc0fc7945

    // before we give extra_kv_pairs away to make the main target,
    // we need to clone deps here for a later use in secondaries.

crates/driver/src/print_build.rs
eed3si9n: Addressed in https://github.com/bazeltools/bzl-gen-build/commit/c97d4b7c8d900f53d63d9e54182d7ffbc0fc7945 by returning Unit.

crates/protobuf_extractor/src/extract_protobuf_imports.rs
eed3si9n: Addressed in https://github.com/bazeltools/bzl-gen-build/commit/c97d4b7c8d900f53d63d9e54182d7ffbc0fc7945 by changing to `?`

crates/protobuf_extractor/src/main.rs
eed3si9n: Addressed in c97d4b7c8d900f53d63d9e54182d7ffbc0fc7945 with the following comment:

        // See https://protobuf.dev/programming-guides/proto3/#importing
        // Protobuf uses file path relative to the workspace as a way of importing:
        //     import "myproject/other_protos.proto";
        // So here we export the relative path as a "definition".

johnynek (approved):
Looks good. Excited to use it.

PR #29, opened 2023-03-29
Add new directive DataRef for users to add in data dependencies.

weichunstevelee:
Added a new directive DataRef in the manual ref family. This allows users to specify data dependencies for build targets. 

The main benefit of having DataRef is it helps bridge the cross language data pass. For example if users have some information built in one language, they are now able to pass through that information through genrules (or other custom rule) to targets in other languages. 

Some other benefits include
1. A natural way for specifying what data to read for python rules.
2. Allow Java/Scala tests to read data from a filepath instead of resources.

### DIFF
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -330,6 +330,10 @@ async fn print_file(
                                 let t = extra_kv_pairs.entry("deps".to_string()).or_default();
                                 t.push(manual_ref.target_value.clone())
                             }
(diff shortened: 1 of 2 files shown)


weichunstevelee:
@ianoc @johnynek

ianoc:
In general i think it would be nice for us to come up with a more general purpose spec to hint at making/extending the targets. but... this is totally in keeping with what we have.

ianoc (approved):

PR #35, opened 2023-04-21
Implement attr.string_list

eed3si9n:
## Problem
There are some situations where we'd want to access the underlying rules, like `scala_library`. We probably don't want to keep adding one-off, language-specific directives.

## Solution
This is somewhat generic solution that passes both the field name and the value.

## Notes
```scala
// bzl_gen_build:attr.string_list: plugins -> { foo }
object SomeMain extends IOApp {
```

This generates

```diff
+    plugins = ["foo"],
```

### DIFF
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -450,7 +450,6 @@ async fn print_file(
 
     apply_manual_refs(&mut extra_kv_pairs, &graph_node.node_metadata);
 
-
     fn append_key_values(
         extra_kv_pairs: &mut HashMap<String, Vec<String>>,
         key: &String,
@@ -461,7 +460,6 @@ async fn print_file(
             .or_default()
             .extend(values.iter().cloned());
     }
-
 
     fn apply_attr_string_lists(
         extra_kv_pairs: &mut HashMap<String, Vec<String>>,
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -300,10 +300,7 @@ async fn print_file(
     };
 
     for (k, lst) in build_config.extra_key_to_list.iter() {
-        extra_kv_pairs
-            .entry(k.clone())
-            .or_default()
-            .extend(lst.iter().cloned())
+        append_key_values(&mut extra_kv_pairs, &k, &lst);
     }
 
     for directive in project_conf
@@ -336,10 +333,7 @@ async fn print_file(
                             }
                         },
                         Directive::AttrStringList(attr) => {
-                            let t = extra_kv_pairs
-                                .entry(attr.attr_name.to_string())
-                                .or_default();
-                            t.push(attr.value.clone())
+                            append_key_values(&mut extra_kv_pairs, &attr.attr_name, &attr.values);
                         }
                     }
                 }
@@ -456,15 +450,25 @@ async fn print_file(
 
     apply_manual_refs(&mut extra_kv_pairs, &graph_node.node_metadata);
 
+
+    fn append_key_values(
+        extra_kv_pairs: &mut HashMap<String, Vec<String>>,
+        key: &String,
+        values: &Vec<String>,
+    ) {
+        extra_kv_pairs
+            .entry(key.clone())
+            .or_default()
+            .extend(values.iter().cloned());
+    }
+
+
     fn apply_attr_string_lists(
         extra_kv_pairs: &mut HashMap<String, Vec<String>>,
         node_metadata: &GraphNodeMetadata,
     ) {
         for attr in node_metadata.attr_string_lists.iter() {
-            let t = extra_kv_pairs
-                .entry(attr.attr_name.to_string())
-                .or_default();
-            t.push(attr.value.clone());
+            append_key_values(extra_kv_pairs, &attr.attr_name, &attr.values);
         }
     }
 
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -335,6 +335,12 @@ async fn print_file(
                                 t.push(manual_ref.target_value.clone())
                             }
                         },
+                        Directive::AttrStringList(attr) => {
+                            let t = extra_kv_pairs
+                                .entry(attr.attr_name.to_string())
+                                .or_default();
+                            t.push(attr.value.clone())
+                        }
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


crates/driver/src/print_build.rs
```diff
@@ -335,6 +335,12 @@ async fn print_file(
                                 t.push(manual_ref.target_value.clone())
                             }
                         },
+                        Directive::AttrStringList(attr) => {
+                            let t = extra_kv_pairs
```
johnynek, line 339: nit: why not chain all the way?
```rust
extra_kv_pairs
  .entry(..
  .or_default()
  .push(attr.value.clone())
```

crates/driver/src/print_build.rs
```diff
@@ -450,6 +456,20 @@ async fn print_file(
 
     apply_manual_refs(&mut extra_kv_pairs, &graph_node.node_metadata);
 
+    fn apply_attr_string_lists(
+        extra_kv_pairs: &mut HashMap<String, Vec<String>>,
+        node_metadata: &GraphNodeMetadata,
+    ) {
+        for attr in node_metadata.attr_string_lists.iter() {
+            let t = extra_kv_pairs
```
johnynek, line 464: same comment as above (also modestly suspicious of the 4 duplicated lines of code. Rust has a great inliner I think, so having a private function you call in both cases may be nice).

crates/shared_types/src/directive.rs
```diff
@@ -464,6 +532,14 @@ mod tests {
             })
         );
 
+        assert_eq!(
+            parse_to_directive("attr.string_list:plugins://x/y/z:artifact"),
```
johnynek, line 536: my only nit here is that using `attr.string_list` it looks like the "type system" used by starlark rules, but actually, these "strings" could also be `label_list`. So I wonder if not using the exact same term would be helpful to avoid confusion.

Maybe just `attr_list` might be better?

crates/shared_types/src/directive.rs
```diff
@@ -136,6 +155,12 @@ pub struct ManualRefConfig {
     pub target_value: String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Deserialize, Serialize)]
+pub struct AttrStringListConfig {
+    pub attr_name: String,
+    pub value: String,
```
johnynek, line 161: shouldn't this be a `Vec<String>` if we have a list? I feel like we should go ahead and support lists if we are claiming it is a list.

crates/shared_types/src/directive.rs
```diff
@@ -276,6 +302,42 @@ impl Directive {
         ))
     }
 
+    fn parse_attr_string_list_directive<'a, E: ParseError<&'a str> + ContextError<&'a str>>(
```
johnynek, line 305: I'd rather support something like `attr_list:foo:["bar", "baz", "quux"]` so we can give the literal list as it will appear in the code. I think it could be as simple as starting with `[` and ending at `]` as long as it is not inside a quote. I know that means you have to parse the strings which is non-trivial... but I mean, having an escape hatch that only works for very narrow cases seems like it just sets us up to revisit.

johnynek (commented):
not completely opposed to merging before we have things working completely, but I'd rather at least have the singleton list case syntactically compatible with the list case.

I worry `attr.string_list:foo:bar` can't be generalized enough so it will make the format complex. Like what if there are multiple values? what if there are spaces in the strings? what if there are new-lines in the strings? etc...

eed3si9n:
> I worry `attr.string_list:foo:bar` can't be generalized enough so it will make the format complex. Like what if there are multiple values? what if there are spaces in the strings? what if there are new-lines in the strings? etc...

Multiple values are supported by having multiple directives:

```scala
// bzl_gen_build:attr.string_list:plugins:foo
// bzl_gen_build:attr.string_list:plugins:bar
```

This generates

```diff
+    plugins = [
+        "bar",
+        "foo",
+    ],
```

eed3si9n:
> not completely opposed to merging before we have things working completely, but I'd rather at least have the singleton list case syntactically compatible with the list case.

You mean like:

```scala
// bzl_gen_build:attr.string_list:plugins:["foo"]
```

? Syntactically this feels like a departure from `ref` and `manual_ref` that we currently have.

crates/shared_types/src/directive.rs
eed3si9n: Yea, the beauty of [`attr`](https://bazel.build/rules/lib/toplevel/attr) is that it's finite, and I did take that name. My thinking is that is `attr.string_list` goes well, we can keep expanding to cover `attr.int` etc. I'd be happy to add `attr.label_list` as a syntactic sugar.

johnynek:
there is a list syntax currently `foo -> {bar, baz}` or something is supporting for adding links. But maybe that was removed since we never used it.

crates/shared_types/src/directive.rs
```diff
@@ -464,6 +532,14 @@ mod tests {
             })
         );
 
+        assert_eq!(
```
johnynek, line 535: can we add a test with multiple string_list given?

I see that it should work, but I think we should test.

johnynek:
maybe worth copying that? I'd don't know...

eed3si9n:
> maybe worth copying that? I'd don't know...

Yea. I'll refactor the function to reuse it for both.

crates/driver/src/print_build.rs
eed3si9n: Refactored to `append_key_values` function in 525096d2c8e7dd0fc2228985b2c42cdb754d76d9

crates/driver/src/print_build.rs
eed3si9n: Refactored to `append_key_values` function in 525096d2c8e7dd0fc2228985b2c42cdb754d76d9

crates/shared_types/src/directive.rs
eed3si9n: Added `attr.label_list` in 525096d2c8e7dd0fc2228985b2c42cdb754d76d9

crates/shared_types/src/directive.rs
eed3si9n: Changed to accept `Vec<String>` in 525096d2c8e7dd0fc2228985b2c42cdb754d76d9

crates/shared_types/src/directive.rs
eed3si9n: Changed to:

```
attr.string_list: plugins -> { //x/y/z:artifact, bar }
```

crates/shared_types/src/directive.rs
eed3si9n: Added a test with two entries:

```rust
        assert_eq!(
            parse_to_directive("attr.string_list: plugins -> { //x/y/z:artifact, bar }"),
....
```

PR #40, opened 2023-04-27
Allow print_build to skip a node if it's not part of the configuration to allow external extracted-defs

weichunstevelee:
This PR updated behavior of print_rs to skip printing instead of returning an error when we see a node that is not included in the built build graph.

**Why do we need this?**
We recently added a way to expose symbols for proto-generated code, and need a way to incorporate that piece of information with what is generated by bzl-gen-build`system-app/extracted-defs`. 

Note: this is probably generally needed for anything crossing languages because they will run at different time (e.g., we have to print build for proto first, and then print build for java, ptyhon, etc). More specifically, imagine the following flow
 
1. print proto build files.
2. extract python symbols for python files, extract python symbols from proto files 
3. print build files for python directories
4. **at this point we can't print build files for proto directories to add py_proto_library rule, as it'd otherwise clean them out**

This can likely be resolved once we are able to do incremental print (i.e., read BUILD and edit it instead of printing a new one). 

**Caveats** 
This also means we silently ignore the graph nodes that are not in the configuration. Though I think it's probably fine as BUILD file printing is rather explicit and if BUILD fails it's obvious? That said I'm also curious if there are better ways, but i'm not familiar with this codebase enough.

### DIFF
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -235,10 +235,8 @@ async fn print_file(
     let (module_config, path_type, _matched_prefix) = if let Some(a) = module_config {
         a
     } else {
-        return Err(anyhow!(
-            "Unable to find any matching configuration for {}",
-            element
-        ));
+        let empty_files = Vec::default();
+        return Ok(empty_files)
     };
     let mut emitted_files = Vec::default();
 


weichunstevelee:
cc: @johnynek @ianoc

crates/driver/src/print_build.rs
```diff
@@ -235,10 +235,8 @@ async fn print_file(
     let (module_config, path_type, _matched_prefix) = if let Some(a) = module_config {
         a
     } else {
-        return Err(anyhow!(
-            "Unable to find any matching configuration for {}",
-            element
-        ));
+        let empty_files = Vec::default();
+        return Ok(empty_files)
```
ianoc, line 239: I think you'd want `return Ok(Default::default());` for idiomatic sense.

ianoc:
I'm a little concerned this would lead to issues in config meaning you'd silently drop write operations -- the more common thing i think from messing this up than getting a build wrong and it yelling at you might be that we would leave old versions of files in place and not update them during re-generation.

When your hitting this today are you hitting that it wants to write out into a proto tree? or what flavor of error?

weichunstevelee:
HI @ianoc, yeah totally agree. I was originally thinking adding one more field saying this is "pure provider", so we only skip when it's pure provider. That sounds like a better approach. 

As for the question - I guess it's the other way around. So we print BUILD files for the proto tree, and when we print BUILD for java or python trees it'll need information from proto tree. If these information is added as part of build graph, the build generator will attempt to print in the proto tree as well. Hence the need for skipping those nodes. 

Do you think it make sense to further mark some nodes as "pure information provider", or provide these additional information through other manner can work here? Also curious if there are other ways

ianoc:
Currently we do have some mechanism to ignore things we get defs for, but without the full setup you have there i don't entirely remember off hand how it works. But we inject things like the external dependency information into the graph.

So either how we inject it (if you look at the script to build the build files it should have a temporary output folder with the json to describe the targets), or the targets are referenced with a `@` at the start, if its that one we could try `@//src/main/proto/...` which would make it look like not a file on disk.

also having an ignore prefix in the config is good with me too

weichunstevelee:
Ah i see. I acutally injected it in the way you mentioned (just not added in graph) but added them in defs. Let me see what i can do. Closing this one for now

ianoc:
Just to be clear what i'm suggesting wouldn't require manually annotating any source files to say they use the type/the dependency. so it would be in the graph. But not in one to be printed. so it would be external. But you may have that already in hand here. Let me know if i can help with other stuff though

PR #45, opened 2023-05-09
Jarscanner Rust port

wrobstory:
Port from Scala to Rust of our jarscanning tool, including the unit test suite. 

I still need to do some testing of the command line tool on some actual jar files, and I'm looking for feedback on how we handle `label_to_allowed_prefixes` from the CLI itself (vs being called as a library, in which case we can pass in the value). 

Given that this PR was already getting lengthy, I wanted to push off integration into bzl-gen-build to a follow-on PR.


crates/jarscanner/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let opt = Opt::parse();
+    let output_path = PathBuf::from(opt.out);
+    let input_jar_path = PathBuf::from(opt.input_jar);
+
+    let label_to_allowed_prefixes = HashMap::new();
+    // Insert any prefixes you want to allow here. For example:
```
wrobstory, line 33: Looking for feedback here: we could make this a CLI argument, or have it read in a JSON file, or?

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::PathBuf;
+use zip::read::ZipArchive;
+
+pub mod errors;
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default, Eq)]
+struct DataBlock {
```
johnynek, line 15: seems weird we need to copy this. Isn't code like this already in another project here? Can we factor so we can share this code across crates?

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    label_or_repo_path: String,
+    data_blocks: Vec<DataBlock>,
+}
+
+fn non_anon(file_name: &str) -> bool {
+    !((file_name.contains("/$") && file_name.contains("$/")) || file_name.contains("$$anon"))
+}
+
+fn not_in_meta(file_name: &str) -> bool {
+    !file_name.starts_with("META-INF")
```
johnynek, line 32: starts with seems risky here. Don't we want `starts_with("META-INF/")` or something? we don't want to have a false positive on `META-INFO` or something.

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    refs: Vec<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default, Eq)]
+pub struct TargetDescriptor {
+    label_or_repo_path: String,
+    data_blocks: Vec<DataBlock>,
+}
+
+fn non_anon(file_name: &str) -> bool {
```
johnynek, line 27: we should add a comment here that this is to filter scala generated classes that users should never reference.

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    match label_to_allowed_prefixes.get(label) {
+        Some(prefix_vec) => classes
+            .into_iter()
+            .filter(|c| prefix_vec.iter().any(|prefix| c.starts_with(prefix)))
+            .collect(),
+        None => classes,
+    }
+}
+
+fn sort_and_deduplicate(vec: &Vec<String>) -> Vec<String> {
```
johnynek, line 90: why can't we take ownership here? Why do we need to clone? Seems like we don't need to clone below.

johnynek, line 90: why can't we take ownership here? Why do we need to clone? Seems like we don't need to clone below.

crates/jarscanner/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+#[command(about = "Jar to containing classes tool", long_about = None)]
+struct Opt {
+    #[arg(long)]
+    out: String,
+
+    #[arg(long)]
+    label: String,
+
+    #[arg(long)]
+    input_jar: String,
```
johnynek, line 20: I think you can use `PathBuf` here or something directly so we don't need to convert from String. Same with the `out` above I think and the relative path below.

crates/jarscanner/src/main.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    #[arg(long)]
+    relative_path: String,
+}
+
+#[tokio::main]
```
johnynek, line 26: a bit weird to use tokio but no other async right? Do we see a path to use async? I'm not sure if you ever would on a single file. If we took a list of files to scan, yes.

crates/jarscanner/src/lib.rs
wrobstory: Ah, yeah it is! A little bit of newbie-ness here. Will fix

crates/jarscanner/src/lib.rs
wrobstory: Ah good question: this is what the current JarScanner does, but I think your suggestion makes sense. Will change.

crates/jarscanner/src/main.rs
wrobstory: Yeah a little bit of pattern matching with what our other CLI tools do here. I'll remove

crates/jarscanner/src/lib.rs
wrobstory: Ah yeah i got in a borrow checker fight elsewhere and ended up here, I think we can sort `classes` directly in the other method

wrobstory:
Ok, this is ready for another pass. Biggest changes: 

1. Addresses review comments
2. I had the CLI take `label_to_allowed_prefixes` as a JSON argument. I realized that if you're calling the CLI from a machine tool, this is probably what you're going to want to do (not write a new file and pass in the path to that file)
3. I added integration tests that round-trip a 2.11 and 2.12 .jar

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+            bzl_gen_build_commands: HashSet::new(),
+        }],
+    })
+}
+
+pub fn emit_result(
+    target_descriptor: &ExtractedData,
+    output_path: &PathBuf,
+) -> Result<(), Box<dyn Error>> {
+    let json = serde_json::to_string_pretty(target_descriptor)?;
```
eed3si9n, line 103: `DataBlock` uses `HashSet`, so wouldn't that produce non-deterministic JSON? Could we add `serialize_with` on `HashSet` fields so it would sort before it serializes to JSON?

wrobstory: Hmmmmm good catch, I had not considered this, but it looks like DataBlack has the serialize_with stuff on all the fields we need: https://github.com/bazeltools/bzl-gen-build/blob/main/crates/shared_types/src/api/extracted_data.rs#L14

Do you think that's sufficient?

eed3si9n: That's great!

eed3si9n (approved):
LGTM on the JSON front

crates/jarscanner/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+    let mut file = File::create(output_path)?;
+    file.write_all(json.as_bytes())?;
+    Ok(())
```
johnynek, line 106: why do `?;` then `Ok(())`? does write_all not return `()`? Can we add a comment so someone isn't tempted to remove the `?` and trailing `Ok`?

johnynek (approved):

crates/jarscanner/src/lib.rs
wrobstory: Here's what the compiler yells at me if I remove the ? and the `Ok(())`: 
```rust
error[E0308]: mismatched types
   --> jarscanner/src/lib.rs:105:5
    |
102 | ) -> Result<(), Box<dyn Error>> {
    |      -------------------------- expected `Result<(), Box<(dyn std::error::Error + 'static)>>` because of return type
...
105 |     file.write_all(json.as_bytes())
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<(), Box<dyn Error>>`, found `Result<(), Error>`
    |
    = note: expected enum `Result<_, Box<(dyn std::error::Error + 'static)>>`
               found enum `Result<_, std::io::Error>`
help: use `?` to coerce and return an appropriate `Err`, and wrap the resulting value in `Ok` so the expression remains of type `Result`
    |
105 |     Ok(file.write_all(json.as_bytes())?)
    |     +++                               ++
```

However, I do think that `Ok(file.write_all(json.as_bytes())?)` is strictly better

wrobstory: ....or is it? Now second guessing myself. I dunno, i agree with you that it feels like I should just be able to call `write_all` and return the Result from it.

wrobstory: I think I'd rather leave what I have than `Ok(file.write_all(json.as_bytes())?)`, having chewed on this for a couple minutes

Issue #42, opened 2023-05-02
Reduce dynamic linking to glibc


eed3si9n:
Now that GitHub stopped Bionic GitHub Actions, the binary are built using Focal (Ubuntu 20.04), which no longer works on Bionic (Ubuntu 18.04), which uses glibc 2.27.

Can this be resolved if we used MUSL?

```bash
cargo build --target x86_64-unknown-linux-musl
```

ianoc:
Yep! using a musl builds at one point was more of a pain, but last i tried this for something it worked super well.

PR #81, opened 2023-07-20
Use musl on Linux

eed3si9n:
Fixes https://github.com/bazeltools/bzl-gen-build/issues/42

## Problem
Bionic vs Jammy situation

## Solution
This switches Linux build to target `x86_64-unknown-linux-musl`


johnynek:
exciting!

PR #89, opened 2023-07-24
Example repo

eed3si9n:
This is an offshoot from https://github.com/bazeltools/bzl-gen-build/pull/13

Problem
-------
In the current state bzl-gen-build is not really usable outside of our own usage since it's relying on custom scripts and macros. This is often necessitated by our tendency to use old version of Python etc.

Solution
--------
This adds an example repository to reimplement some of the scripts and configurations using open source rules_python etc.


crates/python_extractor/src/main.rs
```diff
@@ -123,12 +123,16 @@ async fn main() -> Result<()> {
 }
 
 fn expand_path_to_defs_from_offset(from_given_path: &str, path: &str) -> Vec<String> {
-    if let Some(rem) = path.strip_prefix(from_given_path) {
+    if let Some(rem0) = path.strip_prefix(from_given_path) {
+        let rem = match rem0.strip_prefix("site-packages/") {
```
eed3si9n, line 127: This is a minor change on Python extractor, needed because rules_python Bzlmod support uses pip-tools, which I think places the 3rdparty source files inside a `site-packages/` directory, per module.

ianoc, line 127:
```
if let Some(rem) = path.strip_prefix(from_given_path).and_then(|p| p.strip_prefix("site-packages/").unwrap_or(p) {
```
inner block is ok, too but i'd use `unwrap_or` here. Probably also just put your comment here into the file

ianoc (approved):
one nit, otherwise lgtm

crates/python_extractor/src/main.rs
eed3si9n: Addressed in 456bc96

example/src/main/python/com/example/hello/BUILD.bazel
```diff
@@ -0,0 +1,2 @@
+filegroup(name='hello_files', srcs=glob(include=['**/*.py']), visibility=['//visibility:public'])
+py_library(name='hello', srcs=[':hello_files'], deps=['@@rules_python~0.24.0~pip~pip_39_pandas//:pkg'], visibility=['//visibility:public'])
```
johnynek, line 2: that double `@` looks fishy to me... is that actually correct?

example/src/test/java/com/example/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+java_test(
+    name = "Test1",
```
johnynek, line 2: this is manually written right?

eed3si9n: Yea, I haven't hooked up Java support in this PR.

example/src/main/python/com/example/hello/BUILD.bazel
eed3si9n: Yea Bazel 6's module system seems to use `@@`:

```
$ bazel query @@rules_python~0.24.0~pip~pip_39_pytz//:pkg
@@rules_python~0.24.0~pip~pip_39_pytz//:pkg
```

.github/ci_scripts/integration_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -o errexit  # abort on nonzero exitstatus
+set -o nounset  # abort on unbound variable
+set -o pipefail # don't hide errors within pipes
+
+cd example
+# TOOLING_WORKING_DIRECTORY=/tmp/bzl-gen-build source build_tools/lang_support/create_lang_build_files/regenerate_protos_build_files.sh
```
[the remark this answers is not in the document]
eed3si9n: This `example/` used to be part of the Protobuf PR, but it got too big, so I am splitting it up. I figured Python would be most self-contained (without jar scanner etc) to demonstrate, but eventually we should try to hook it up as well using the new jar scanner.

.github/ci_scripts/integration_test.sh
```diff
@@ -0,0 +1,10 @@
...
+TOOLING_WORKING_DIRECTORY=/tmp/bzl-gen-build source build_tools/lang_support/create_lang_build_files/regenerate_python_build_files.sh
```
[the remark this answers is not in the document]
eed3si9n: Addressed in cdec70c70f4aab9610c547ffd076de0a15a3d92d

johnynek (approved):

PR #110, opened 2023-08-16
Add jarscanner to the release

wrobstory:
When looking at the released packages, I realized that I forgot to add jarscanner to the crates we export for release.


eed3si9n (approved):
Thanks @wrobstory!

Issue #114, opened 2023-08-17
Support multiple targets in a directory (1:1:n setup)


eed3si9n:
## Steps

Given a directory named `utils`, currently bzl-gen-build generates:

```python
filegroup(
    name = "utils_files",
    srcs = glob(include = ["**/*.proto"]),
    visibility = ["//visibility:public"],
)

proto_library(
    name = "utils",
    srcs = [":utils_files"],
    visibility = ["//visibility:public"],
    deps = [],
)
```

## Problem

While grabbing all source files in the directory works (1:1:1) for languages like Java, Scala, Python etc, it might not be desirable for Protobuf. See for example Google's `src/google/protobuf` targets - https://github.com/protocolbuffers/protobuf/blob/v24.0/src/google/protobuf/BUILD.bazel

```python
proto_library(
    name = "struct_proto",
    srcs = ["struct.proto"],
    strip_import_prefix = "/src",
)

proto_library(
    name = "timestamp_proto",
    srcs = ["timestamp.proto"],
    strip_import_prefix = "/src",
)

proto_library(
    name = "type_proto",
    srcs = ["type.proto"],
    strip_import_prefix = "/src",
    deps = [
        "//:any_proto",
        "//:source_context_proto",
    ],
)
....
```

## Expectations

Similarly, we should support 1:1:n-setup where one `proto_library(...)` is created per `.proto` file.

PR #117, opened 2023-08-18
Implement --no-aggregate-source mode

eed3si9n:
Fixes https://github.com/bazeltools/bzl-gen-build/issues/114

Problem
-------
For protobuf, we'd like to generate one target per `*.proto` file,
not aggregate all `*.proto` under the directory.

See for example Google's `src/google/protobuf` targets - https://github.com/protocolbuffers/protobuf/blob/v24.0/src/google/protobuf/BUILD.bazel

Solution
--------
This implements command-line option called `--no-aggregrate-source`
which change the way graph is constructed to use files, instead of
the directories.


crates/driver/src/print_build.rs
```diff
@@ -921,60 +970,85 @@ proto_library(
...
+            6,
             r#"load('@com_google_protobuf//:protobuf.bzl', 'py_proto_library')
 load('@rules_proto//proto:defs.bzl', 'proto_library')
 
-filegroup(
-    name='protos_files',
-    srcs=glob(include=['**/*.proto']),
+proto_library(
+    name='a_proto',
+    srcs=['a.proto'],
```
eed3si9n, line 1000: See that `proto_library(...)` is generated one per source file.

crates/protobuf_extractor/src/extract_protobuf_imports.rs
```diff
@@ -3,44 +3,80 @@ use tree_sitter::Parser;
...
+                        quoted.next_back();
+                        if !(quoted.as_str().starts_with("google/")) {
                             buf.push(quoted.as_str().to_string());
                         }
                     }
-                _        => ()
+                }
+                "comment" => {
+                    let raw_comment = child_node.utf8_text(bytes)?.to_string();
+                    if let Some(comment_line) = raw_comment.trim_start().strip_prefix("//") {
```
johnynek, line 38: I am not in love with the nesting. What about something like:
```rust
let opt_cmd = raw_comment
  .trim_start()
  .strip_prefix("//")
  .and_then(|comment_line| comment_line.trim_start().strip_prefix("bzl_gen_build"))
  .and_then(|matching| matching.trim_start().strip_prefix(':'));
 
if let Some(bzl_command) = opt_cmd {
  bzl_gen_build_commands.push(bzl_command.trim().to_string());
}
```

crates/driver/src/print_build.rs
```diff
@@ -248,127 +248,271 @@ where
...
 
     let target_folder = opt.working_directory.join(&element);
-    let target_name = target_folder
-        .file_name()
-        .unwrap()
-        .to_str()
-        .unwrap()
-        .to_string();
+    let base_name = to_file_name(&target_folder);
+    let mut t = TargetEntries {
```
johnynek, line 252: what about adding `derive(Default)` to `TargetEntries` so this is `Default::default()`

crates/driver/src/print_build.rs
```diff
@@ -248,127 +248,271 @@ where
...
+        let node_file = opt.working_directory.join(&graph_node.node_label);
+        let node_file_name = to_file_name(&node_file);
+        let target_name = if !opt.no_aggregate_source {
+            base_name.clone()
+        } else {
+            node_file_name.replace(".", "_")
+        };
+        let mut extra_kv_pairs: HashMap<String, Vec<String>> = HashMap::default();
+
+        if !graph_node.dependencies.is_empty() {
```
johnynek, line 265: nit: can we reduce repetition:
```rust
fn add_non_empty(key: String, labels: Vector<String>, mut& extra_kv_pairs: Hash<String, Vector<String>>) {
  if !labels.is_empty() {
    let vals = labels.iter()
                .map(|e| to_label(&opt, e))
                .collect();
    extra_kv_pairs.insert(key, vals)
  }
}
```
then call it twice?

example/src/main/protos/com/example/foo/BUILD.bazel
```diff
@@ -1,8 +1,8 @@
 load('@rules_proto//proto:defs.bzl', 'proto_library')
 load('@rules_python//python:proto.bzl', 'py_proto_library')
-proto_library(name='c_proto', srcs=['c.proto'], visibility=['//visibility:public'])
-java_proto_library(name='c_proto_java', deps=[':c_proto'], visibility=['//visibility:public'])
-py_proto_library(name='c_proto_py', deps=[':c_proto'], visibility=['//visibility:public'])
-proto_library(name='b_proto', srcs=['b.proto'], visibility=['//visibility:public'])
+proto_library(name='b_proto', srcs=['b.proto'], deps=['//src/main/protos/com/example:a_proto'], visibility=['//visibility:public'])
 java_proto_library(name='b_proto_java', deps=[':b_proto'], visibility=['//visibility:public'])
```
weichunstevelee, line 4: one nit-picking is that can we rename these to `c_java_proto` to be consistent with convention?

weichunstevelee (commented):
Thanks, this is really exciting!

example/src/main/protos/com/example/foo/BUILD.bazel
eed3si9n: Currently `c_proto` is generated from `c.proto`, and then `_java` gets appended because it's a secondary rule off of `proto_library`. I don't know if we can change that easily.

crates/driver/src/extract_defrefs.rs
```diff
@@ -194,24 +196,30 @@ async fn async_extract_def_refs(
 async fn merge_defrefs(
     concurrent_io_operations: &Semaphore,
     path_sha_to_merged_defrefs: &'static Path,
-    directory: String,
+    entry: String,
     project_conf: &'static ProjectConf,
     mut work_items: Vec<ProcessedFile>,
     sha_of_conf_config: Arc<String>,
+    no_aggregate_source: bool,
```
johnynek, line 203: this worries me. In the case of proto this makes sense because protos are acyclic. But many programming languages (such as scala) are not enforced to be acyclic. Furthermore: this option can conflict with the def/ref graph (if a -> b and b -> a but no_aggregate_source == true, what do we do?).

crates/driver/src/extract_defrefs.rs
```diff
@@ -194,24 +196,30 @@ async fn async_extract_def_refs(
...
     work_items.sort_by(|a, b| a.sha256.cmp(&b.sha256));
 
     let merged_sha = Sha256Value::hash_iter_bytes(
         work_items
             .iter()
             .map(|e| e.sha256.as_bytes())
-            .chain(std::iter::once(sha_of_conf_config.as_bytes())),
+            .chain(std::iter::once(sha_of_conf_config.as_bytes()))
+            .chain(std::iter::once(if no_aggregate_source {
+                &b"1"[0..1]
```
johnynek, line 213: nit: this looks like we are using ascii `"1"` for true, rather than the byte with value 1. Am I understanding that correctly? I'd personally rather use byte == 0 and byte == 1 not the ascii.

crates/driver/src/extract_defs.rs
```diff
@@ -74,18 +74,16 @@ pub async fn extract_exports(
...
-            u_k.rfind('/').map(|e| u_k.split_at(e).0).unwrap_or(u_k)
+    for (rel_path, content_path) in extracted_mappings.relative_path_to_extractmapping.iter() {
+        let entry = if !opt.no_aggregate_source {
+            to_directory(&rel_path)
         } else {
-            k.as_str()
+            rel_path.to_string()
         };
-        if let Some(v) = work.get_mut(directory) {
+        if let Some(v) = work.get_mut(&entry) {
```
johnynek, line 83: nit: this entry to directory rename could be in a separate PR to make this diff smaller and easier to review.

crates/driver/src/main.rs
```diff
@@ -87,6 +87,9 @@ pub struct Opt {
     #[clap(long)]
     cache_path: PathBuf,
 
+    #[clap(long)]
+    no_aggregate_source: bool,
```
johnynek, line 91: can we add a comment as to what this means? Is it minimize the number of targets or error if you cannot make a DAG with this file in a single node?

crates/driver/src/main.rs
```diff
@@ -181,6 +184,14 @@ fn read_all_project_conf(input_path: &Path, working_directory: &Path) -> Result<
     Ok(v)
 }
 
+pub fn to_directory(rel_path: &String) -> String {
```
johnynek, line 187: nit: I think we can make this `(rel_path: String) -> String` and avoid the clone. It looks like our use case can handle taking ownership. Also, since we are unconditionally calling clone, if the caller only has a `&str` or `&String` they can call `.clone()` and recover the efficiency of this implementation.

crates/protobuf_extractor/src/extract_protobuf_imports.rs
eed3si9n: Addressed in https://github.com/bazeltools/bzl-gen-build/pull/117/commits/e034038a27cfab319216bf41118a33c2e2fb8dce

crates/driver/src/extract_defrefs.rs
eed3si9n: For now I am planning on enabling this only for Protobufs.

> (if a -> b and b -> a but no_aggregate_source == true, what do we do?).

I'd say we should fix the loop instead of the generator.

crates/driver/src/extract_defrefs.rs
eed3si9n: Yea, you're probably right.

crates/driver/src/main.rs
eed3si9n: Addressed in aa1a4a68b019b67a2cd9086ac678cd4408877ff6

crates/driver/src/extract_defrefs.rs
eed3si9n: Addressed in https://github.com/bazeltools/bzl-gen-build/pull/117/commits/312fd34366f893136280e71bfe20d30c2b4f87f2

crates/driver/src/print_build.rs
eed3si9n: Addressed in https://github.com/bazeltools/bzl-gen-build/pull/117/commits/b81b4341e7af6afbe90894aa28e682cdcc3022d4

crates/driver/src/print_build.rs
eed3si9n: Addressed in 662f6714d7713bb5d5b3d29da9cc3286b97ff33c

crates/driver/src/main.rs
eed3si9n: Addressed in abefc9531c9940200281fa63f42e3a35011d47f0

crates/protobuf_extractor/src/extract_protobuf_imports.rs
eed3si9n: I've split directive related things into https://github.com/bazeltools/bzl-gen-build/pull/118

crates/driver/src/extract_defs.rs
eed3si9n: Here's a PR to split out some of the directory/entry renames - https://github.com/bazeltools/bzl-gen-build/pull/122

eed3si9n:
Squashed commits, and rebased on top of main.

wrobstory (approved):

PR #118, opened 2023-08-18
Implement directives support for Protobuf

eed3si9n:
This is an offshoot of https://github.com/bazeltools/bzl-gen-build/pull/117

Problem/Solution
----------------
Currently Protobuf extractor doesn't support directives.
This adds it via comment similar to other languages.


johnynek (approved):
Thanks for splitting.

PR #123, opened 2023-08-24
Dont publish BUILD or WORKSPACE in the published packages

wrobstory:
Ian do you remember why these were here in the first place (and most importantly, if they were necessary/needed for any OSS stuff)?

### DIFF
--- a/prepare_all_apps.sh
+++ b/prepare_all_apps.sh
@@ -44,11 +44,4 @@ cp ${RUST_TARGET_DIR}/bzl_gen_protobuf_extractor $PREPARE_ALL_OUTPUT_DIR/protos-
 rm -f  $PREPARE_ALL_OUTPUT_DIR/jarscanner || true
 cp ${RUST_TARGET_DIR}/bzl_gen_jarscanner $PREPARE_ALL_OUTPUT_DIR/jarscanner
 
-echo "workspace(name = \"external_build_tooling_gen\")" > $PREPARE_ALL_OUTPUT_DIR/WORKSPACE
-
-cat > $PREPARE_ALL_OUTPUT_DIR/BUILD.bazel <<EOF
-exports_files(["WORKSPACE", "python-entity-extractor"])
-EOF
-
-
 echo "wrote all outputs to $PREPARE_ALL_OUTPUT_DIR" 1>&2


ianoc:
It may not care about it anymore, but used to matter when downloading an external repository to have the workspace/and a build file. You could put in ones in the download path/utils in the destination repo or you took it from the source. This made this not have some warnings/work nicely at some point. May not be valid now though

ianoc:
(No workspace in the dep, used to be some sort of warning printed out iirc)

eed3si9n:
So it seems like the way external_build_tooling_gen is set up currently assumes it to be a subworkspace. Doesn't necessarily mean we need to ship `WORKSPACE` to tgz, but we'd need to adjust something:

> `ERROR: /home/runner/.cache/bazel/_bazel_runner/205dc0b8cad023dc197a399effce7489/external/rules_python~0.24.0~pip~pip_39_python_dateutil/BUILD.bazel:22:11: @rules_python~0.24.0~pip~pip_39_python_dateutil//:pkg depends on @external_build_tooling_gen//:python-entity-extractor in repository @external_build_tooling_gen which failed to fetch. no such package '@external_build_tooling_gen//': No WORKSPACE file found in /home/runner/.cache/bazel/_bazel_runner/205dc0b8cad023dc197a399effce7489/external/external_build_tooling_gen`
> `ERROR: Analysis of aspects '[//build_tools/bazel_rules/wheel_scanner:wheel_scanner.bzl%wheel_scanner_aspect] with parameters {} on @rules_python~0.24.0~pip~pip//python_dateutil:pkg' failed; build aborted: Loading failed`

eed3si9n:
See https://github.com/bazeltools/bzl-gen-build/blob/6e76646d0e4b571c5e25c5ee2bd29b0748751b9a/example/build_tools/bazel_rules/wheel_scanner/py_build_commands.py#L26-L30

It seems like this uses `--override_repository` trick to allow configuring the bzl-gen-build adjacent tooling (like wheel scanner) directory to be configurable.

ianoc:
State/knowledge may be stale, but

All external dependencies are repositories, or were in bazel.  They all eventually have a WORKSPACE, you can use the macros in bazel to fetch a http dependency and add the workspace in. I think generally there is always a BUILD file too, but its permissable not to have one, it can just cause things not to be easily accessible. http_archive can inject one too. 
http archive calls through to https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/utils.bzl#L60-L65

i'm guessing this is failing because `local_repository` isn't trying to inject new ones in, we could try use https://bazel.build/reference/be/workspace#new_local_repository but it might change relative path mappings possibly since it does a symlink against an external folder in a local temp path created afaik

Why are we are trying to delete these?

ianoc:
I think override repository will probably fail too yeah since its not a valid repository without the workspace file at least

wrobstory:
> Why are we are trying to delete these?

I was trying to point to these as an external tool, but I'm looking at what Eugene posted above and the `--override_repository=external_build_tooling_gen=${{BZL_GEN_BUILD_TOOLS_PATH}}--` was actually the thing I should have been reaching for instead. 

Will close, thanks for taking a look, hope you're doing well 😄

PR #134, opened 2023-08-29
Attempt to improve jarscanner perf

wrobstory:
Tried a few tricks aaaaaaand: 
```
---- Timing for 50 runs ----
        4.37 real         1.26 user         2.91 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.26 real         1.24 user         2.88 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.27 real         1.25 user         2.90 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.21 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.22 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.20 real         1.21 user         2.87 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
```

Didn't really see much improvement 😭 Totally open to suggestions on other things to change here, if you see errors in my little perf testing script, etc.

### DIFF
--- a/crates/jarscanner/src/lib.rs
+++ b/crates/jarscanner/src/lib.rs
@@ -1,6 +1,6 @@
 use crate::errors::JarscannerError;
 
-use std::collections::{HashMap, HashSet, BTreeSet};
+use std::collections::{BTreeSet, HashMap, HashSet};
 
 use std::fs::File;
 use std::io::{BufReader, BufWriter};
@@ -24,10 +24,7 @@ fn ends_in_class(file_name: &str) -> bool {
     file_name.ends_with(".class")
 }
 
-fn file_name_to_class_names(
-    file_name_ref: &str,
-    result: &mut BTreeSet<String>,
-) {
+fn file_name_to_class_names(file_name_ref: &str, result: &mut BTreeSet<String>) {
     if non_anon(file_name_ref) && not_in_meta(file_name_ref) && ends_in_class(file_name_ref) {
         let length_of_name = file_name_ref.len();
         let mut file_name_res = String::with_capacity(length_of_name);
@@ -38,7 +35,6 @@ fn file_name_to_class_names(
         // We know this ends in .class, so we're going to need to track the index of the last 6 chars
         let end_idx = length_of_name - 6;
         for file_char in file_name_ref.chars() {
-
             if idx == end_idx {
                 break;
             }
@@ -94,8 +90,10 @@ fn filter_prefixes(
     label_to_allowed_prefixes: &HashMap<String, Vec<String>>,
 ) {
     match label_to_allowed_prefixes.get(label) {
-        Some(prefix_set) => classes.retain(|c| prefix_set.iter().any(|prefix| c.starts_with(prefix))),
-        None => ()
+        Some(prefix_set) => {
+            classes.retain(|c| prefix_set.iter().any(|prefix| c.starts_with(prefix)))
+        }
+        None => (),
     }
 }
 
@@ -209,10 +207,7 @@ mod tests {
         let expected = classes.clone();
 
         filter_prefixes("foo", &mut classes, &label_to_allowed_prefixes);
-        assert_eq!(
-            classes,
-            expected
-        );
+        assert_eq!(classes, expected);
 
         let mut filtered_classes = BTreeSet::new();
         filtered_classes.insert("com.netflix.iceberg.Foo".to_string());
@@ -224,11 +219,8 @@ mod tests {
         filter_prefixes(
             "@jvm__com_netflix_iceberg__bdp_iceberg_spark_2_12//:jar",
             &mut filtered_classes,
-            &label_to_allowed_prefixes
+            &label_to_allowed_prefixes,
         );
-        assert_eq!(
-            filtered_classes,
-            expected
-        );
+        assert_eq!(filtered_classes, expected);
     }
 }
--- a/crates/jarscanner/src/lib.rs
+++ b/crates/jarscanner/src/lib.rs
@@ -4,7 +4,8 @@ use crate::errors::JarscannerError;
 use std::collections::{HashMap, HashSet};
 
 use std::fs::File;
-use std::io::prelude::*;
+use std::io::BufWriter;
+use std::io::Write;
 use std::path::PathBuf;
 use zip::read::ZipArchive;
 
@@ -105,8 +106,9 @@ pub fn emit_result(
     output_path: &PathBuf,
 ) -> Result<(), JarscannerError> {
     let json = serde_json::to_string_pretty(target_descriptor)?;
-    let mut file = File::create(output_path)?;
-    file.write_all(json.as_bytes())?;
+    let file = File::create(output_path)?;
+    let mut writer = BufWriter::new(file);
+    writer.write_all(json.as_bytes())?;
     Ok(())
 }
 
--- a/crates/jarscanner/src/lib.rs
+++ b/crates/jarscanner/src/lib.rs
@@ -1,4 +1,5 @@
 use crate::errors::FileNameError;
+use crate::errors::JarscannerError;
 
 use std::collections::{HashMap, HashSet};
 use std::error::Error;
@@ -47,7 +48,7 @@ fn file_name_to_class_names(file_name: &str) -> Result<Vec<String>, FileNameErro
     }
 }
 
-fn read_zip_archive(input_jar: &PathBuf) -> Result<HashSet<String>, Box<dyn Error>> {
+fn read_zip_archive(input_jar: &PathBuf) -> Result<HashSet<String>, JarscannerError> {
     let file = File::open(input_jar)?;
     let archive = ZipArchive::new(file)?;
 
@@ -55,7 +56,7 @@ fn read_zip_archive(input_jar: &PathBuf) -> Result<HashSet<String>, Box<dyn Erro
     for file_name in archive.file_names() {
         match file_name_to_class_names(file_name) {
             Ok(class_names) => result.extend(class_names.into_iter()),
-            Err(err) => return Err(Box::new(err)),
+            Err(err) => return Err(JarscannerError::from(err))
         }
     }
 
@@ -81,7 +82,7 @@ pub fn process_input(
     input_jar: &PathBuf,
     relative_path: &str,
     label_to_allowed_prefixes: &HashMap<String, Vec<String>>,
-) -> Result<ExtractedData, Box<dyn Error>> {
+) -> Result<ExtractedData, JarscannerError> {
     let raw_classes = read_zip_archive(input_jar)?;
     let classes = filter_prefixes(label, raw_classes, &label_to_allowed_prefixes);
 
@@ -99,7 +100,7 @@ pub fn process_input(
 pub fn emit_result(
     target_descriptor: &ExtractedData,
     output_path: &PathBuf,
-) -> Result<(), Box<dyn Error>> {
+) -> Result<(), JarscannerError> {
     let json = serde_json::to_string_pretty(target_descriptor)?;
     let mut file = File::create(output_path)?;
     file.write_all(json.as_bytes())?;
--- a/crates/jarscanner/src/lib.rs
+++ b/crates/jarscanner/src/lib.rs
@@ -1,9 +1,8 @@
 use crate::errors::JarscannerError;
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, BTreeSet};
 
 use std::fs::File;
-use std::io::Write;
 use std::io::{BufReader, BufWriter};
 use std::path::PathBuf;
 use zip::read::ZipArchive;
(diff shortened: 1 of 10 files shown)


crates/jarscanner/src/lib.rs
```diff
@@ -38,27 +42,32 @@ fn file_name_to_class_names(file_name: &str) -> Result<Vec<String>, FileNameErro
             .replace(r"/", ".");
```
eed3si9n, line 42: These are regex right? I wonder if this is becoming the hotspot.

eed3si9n: A potential optimization might be to use regex crate, and move the regex compilation out of the loop, so we perform it once per expression?

wrobstory: That's a good idea, I'll give it a go

wrobstory: Take a look at my latest commit, these are the results:
```
---- Timing for 50 runs ----
        4.37 real         1.26 user         2.91 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.26 real         1.24 user         2.88 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.27 real         1.25 user         2.90 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.21 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.22 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.20 real         1.21 user         2.87 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.58 real         1.28 user         3.09 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
---- Timing for 50 runs ----
        4.52 real         1.28 user         3.10 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
---- Timing for 50 runs ----
        4.53 real         1.29 user         3.11 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
``` 

Honestly quite surprised here.

wrobstory:
Latest commit looks like this: 
```
---- Timing for 50 runs ----
        4.37 real         1.26 user         2.91 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.26 real         1.24 user         2.88 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.27 real         1.25 user         2.90 sys
Git Commit: 75e510346d1ab50b53003cc1eb6e4b247532bdf3
---------------------------
---- Timing for 50 runs ----
        4.21 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.22 real         1.22 user         2.89 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.20 real         1.21 user         2.87 sys
Git Commit: fd73ac001898a17ede1e584d98301868dc877160
---------------------------
---- Timing for 50 runs ----
        4.58 real         1.28 user         3.09 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
---- Timing for 50 runs ----
        4.52 real         1.28 user         3.10 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
---- Timing for 50 runs ----
        4.53 real         1.29 user         3.11 sys
Git Commit: 30ffea59861f4c8db85e86ea4131b78b5bf9b8c0
---------------------------
---- Timing for 50 runs ----
        0.98 real         0.49 user         0.30 sys
Git Commit: 4fb258ec381adc133e878e8268cdb219bfc6a45e
---------------------------
---- Timing for 50 runs ----
        0.77 real         0.37 user         0.24 sys
Git Commit: 627348f6ea4dd8187b6d4dafe56e5af71c436a08
---------------------------
---- Timing for 50 runs ----
        0.71 real         0.37 user         0.25 sys
Git Commit: 627348f6ea4dd8187b6d4dafe56e5af71c436a08
---------------------------
---- Timing for 50 runs ----
        0.71 real         0.37 user         0.25 sys
Git Commit: 627348f6ea4dd8187b6d4dafe56e5af71c436a08
---------------------------
```

wrobstory:
Ok this is def ready for review/merge now

crates/jarscanner/src/lib.rs
```diff
@@ -25,40 +26,71 @@ fn ends_in_class(file_name: &str) -> bool {
...
+            buffer.push(file_name_res.replace(".package", ""));
+            Ok(())
         } else {
-            Ok(vec![dotted])
+            buffer.push(file_name_res);
+            Ok(())
         }
     } else {
-        Ok(vec![])
+        Ok(())
```
johnynek, line 75: nit: looks like all branches return `Ok(())` can we move that to the very bottom and return once?

crates/jarscanner/src/lib.rs
```diff
@@ -25,40 +26,71 @@ fn ends_in_class(file_name: &str) -> bool {
...
-            .replace(r"/$", r"/")
-            .replace("$", ".")
-            .replace(r"/", ".");
-        let replace_pkg = dotted.replace(".package", "");
-        if dotted.contains(".package") {
-            Ok(vec![dotted, replace_pkg])
+fn file_name_to_class_names(
+    file_name_ref: &str,
+    buffer: &mut Vec<String>,
+) -> Result<(), FileNameError> {
```
johnynek, line 32: I don't see that we actually ever return `Err`. Can this return `()` now? If so, do we still need `FileNameError`?

crates/jarscanner/src/lib.rs
```diff
@@ -25,40 +26,71 @@ fn ends_in_class(file_name: &str) -> bool {
...
 
-fn read_zip_archive(input_jar: &PathBuf) -> Result<HashSet<String>, Box<dyn Error>> {
+fn read_zip_archive(input_jar: &PathBuf) -> Result<HashSet<String>, JarscannerError> {
     let file = File::open(input_jar)?;
-    let archive = ZipArchive::new(file)?;
+    let reader = BufReader::with_capacity(32000, file);
+    let archive = ZipArchive::new(reader)?;
 
     let mut result = HashSet::new();
+    let mut buf = Vec::new();
```
johnynek, line 85: why do we need to add to a Vec then add those to a HashSet? Can't we just pass the `&mut` of the `HashSet` in instead?

crates/jarscanner/src/lib.rs
```diff
@@ -99,10 +131,11 @@ pub fn process_input(
 pub fn emit_result(
     target_descriptor: &ExtractedData,
     output_path: &PathBuf,
-) -> Result<(), Box<dyn Error>> {
+) -> Result<(), JarscannerError> {
     let json = serde_json::to_string_pretty(target_descriptor)?;
-    let mut file = File::create(output_path)?;
-    file.write_all(json.as_bytes())?;
+    let file = File::create(output_path)?;
+    let mut writer = BufWriter::new(file);
```
johnynek, line 137: this defaults to 8kb: https://doc.rust-lang.org/std/io/struct.BufWriter.html#method.new

maybe instead use `with_capacity` so we can see how big and maybe use something bigger like `32k` or something?

crates/jarscanner/src/lib.rs
```diff
@@ -25,40 +26,71 @@ fn ends_in_class(file_name: &str) -> bool {
...
+                    was_slash = false;
+                    file_name_res.push(file_char)
+                }
+            }
+            idx += 1;
+        }
+
+        if file_name_res.contains(".package") {
+            buffer.push(file_name_res.clone());
+            buffer.push(file_name_res.replace(".package", ""));
```
johnynek, line 68: if we move this first we can avoid the clone:

```rust
// replace takes a ref and returns a new String
buffer.push(file_name_res.replace(".package", ""));
// now we can consume the String and not clone
buffer.push(file_name_res);    
```

crates/jarscanner/src/lib.rs
```diff
@@ -25,40 +26,71 @@ fn ends_in_class(file_name: &str) -> bool {
...
+                }
+                _ => {
+                    was_slash = false;
+                    file_name_res.push(file_char)
+                }
+            }
+            idx += 1;
+        }
+
+        if file_name_res.contains(".package") {
```
johnynek, line 66: idea: according to this: https://www3.nd.edu/~busiforc/handouts/cryptography/letterfrequencies.html `k` is the least common english character. We can't match `.package` if we don't have at least one `k`. So a cheap optimization would be above to track if we have `k`:

```rust
let mut has_k = false;
///
                _ => {
                    was_slash = false;
                    // track if we have a k, and if we don't we definitely don't contain "package" as a substring.
                    has_k |= file_char == 'k';
                    file_name_res.push(file_char)
                }
  if has_k && file_name_res.contains(".package") {
```
then we should be able to skip the contains call in most cases which seems to be coming up enough to matter in the performance.

crates/jarscanner/src/lib.rs
```diff
@@ -99,10 +131,11 @@ pub fn process_input(
 pub fn emit_result(
     target_descriptor: &ExtractedData,
     output_path: &PathBuf,
-) -> Result<(), Box<dyn Error>> {
+) -> Result<(), JarscannerError> {
     let json = serde_json::to_string_pretty(target_descriptor)?;
```
johnynek, line 135: https://docs.rs/serde_json/latest/serde_json/fn.to_writer_pretty.html

how about instead don't build the `json`:
```rust
pub fn emit_result(
    target_descriptor: &ExtractedData,
    output_path: &PathBuf,
) -> Result<(), JarscannerError> {
    let file = File::create(output_path)?;
    let writer = BufWriter::new(file);
    serde_json::to_writer_pretty(writer, target_descriptor)?;
    Ok(())
 }
```
this may skip a copy of the final output which can also be a performance win.

crates/jarscanner/src/lib.rs
johnynek: Of course if we want to keep the order the same we can just capture the replace in a let and then push in the same order.

crates/jarscanner/src/lib.rs
```diff
@@ -25,54 +24,76 @@ fn ends_in_class(file_name: &str) -> bool {
...
+            match file_char {
+                '/' => {
+                    was_slash = true;
+                    file_name_res.push('.')
+                }
+                '$' => {
+                    if idx == end_idx - 1 {
+                        // Do nothing
+                    } else if !was_slash {
+                        was_slash = false;
```
johnynek, line 51: I think this is a bug.

```rust
if !was_slash
```
implies `was_slash == false` and we reset it to false.

I think we need to write a test that fails assuming my claim is true that the algo is wrong, then fix it with:
```rust
if !was_slash {
  // we know was_slash == false in this branch
  file_name_res.push('.')
}
else {
  // we previously saw /
  was_slash = false;
}
```
I think we really want to change the code to track `was_dot` and just don't emit two adjacent `.` characters. So we would treat `/` and `$` the same in that case.

johnynek (commented):
one minor issue still I think.

wrobstory:
@johnynek I think its possible our in-house scala impl has a bug: `scala/package$$hash$colon$colon$.class` gets turned into `scala.collection.package..colon.plus`. I added a test case here to cover it.

crates/jarscanner/src/lib.rs
```diff
@@ -192,6 +192,29 @@ mod tests {
...
         assert_eq!(vec_expected, vec!["scala.runtime", "scala.runtime.package"]);
+
+        // We should handle the /$ pattern correctly
+        set.clear();
+        file_name_to_class_names("scala/collection/immutable/$colon$colon$.class", &mut set);
+        let mut vec_expected = set.iter().collect::<Vec<_>>();
+        vec_expected.sort();
+        assert_eq!(
+            vec_expected,
+            vec!["scala.collection.immutable.colon.colon",]
```
johnynek, line 203: this doesn't seem quite right either and brings up a design flaw in the current system: in scala the name of this symbol would be `::` but that's not how it gets encoded in the jar, or for java... so java could refer to it as `$colon$colon` but scala wouldn't.

We don't need to fix this here, but it is an issue with the system: we need to translate symbol names to some universal encoding, but that's not totally obvious.

johnynek (approved):

PR #141, opened 2023-09-15
Make reader tolerant to files without utf-8 encoding, add some tests

wrobstory:
When we tried to add IPython as a dependency, the python_extractor blew up trying to read the nonascii.py file that's now part of the test cases in this PR. 

I had to do a bunch of refactoring to make this more testable- the actual code changes aren't that big, but I did need to move some things into lib.rs. I'll note below where my actual changes are.

### DIFF
--- a/crates/python_extractor/src/lib.rs
+++ b/crates/python_extractor/src/lib.rs
@@ -1,0 +1,183 @@
+use std::{
+    collections::{BTreeSet, HashSet},
+    io::Read,
+    path::PathBuf,
+};
+
+use bzl_gen_build_python_utilities::PythonProgram;
+
+use bzl_gen_build_shared_types::api::extracted_data::{DataBlock, ExtractedData};
+
+use anyhow::{Context, Result};
+
+mod extract_py_bzl_gen_build_commands;
+mod extract_py_imports;
+
+pub async fn extract_python(
+    relative_input_paths: String,
+    working_directory: PathBuf,
+    output: PathBuf,
+    label_or_repo_path: String,
+    disable_ref_generation: bool,
+    import_path_relative_from: Option<String>,
+) -> Result<()> {
+    let mut relative_input_paths: Vec<String> =
+        if let Some(suffix) = relative_input_paths.strip_prefix('@') {
+            std::fs::read_to_string(PathBuf::from(suffix))?
+                .lines()
+                .map(|e| e.to_string())
+                .collect()
+        } else {
+            vec![relative_input_paths.clone()]
+        };
+
+    relative_input_paths.sort();
+
+    let mut data_blocks: Vec<DataBlock> = Default::default();
+
+    for relative_path in relative_input_paths {
+        let input_file = working_directory.join(&relative_path);
+        let mut refs: HashSet<String> = Default::default();
+        let mut defs: BTreeSet<String> = Default::default();
+        let mut bzl_gen_build_commands: HashSet<String> = Default::default();
+
+        let mut file = std::fs::File::open(&input_file).with_context(|| {
+            format!(
+                "While attempting to open file: {:?
+    }",
+                input_file
+            )
+        })?;
+        let mut buffer = Vec::new();
+        file.read_to_end(&mut buffer).with_context(|| {
+            format!(
+                "While attempting to read file: {:?
+    }",
+                input_file
+            )
+        })?;
+
+        let input_str = String::from_utf8_lossy(&buffer).into_owned();
+
+        bzl_gen_build_commands.extend(extract_py_bzl_gen_build_commands::extract(
+            input_str.as_str(),
+        ));
+
+        if !disable_ref_generation {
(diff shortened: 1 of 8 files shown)


crates/python_extractor/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+
+    let mut data_blocks: Vec<DataBlock> = Default::default();
+
+    for relative_path in relative_input_paths {
+        let input_file = working_directory.join(&relative_path);
+        let mut refs: HashSet<String> = Default::default();
+        let mut defs: BTreeSet<String> = Default::default();
+        let mut bzl_gen_build_commands: HashSet<String> = Default::default();
+
+        let mut file = std::fs::File::open(&input_file).with_context(|| {
```
wrobstory, line 44: From this line down to line 60 are the only real changes in here. Notably, it now uses `from_utf8_lossy` which will replace non-utf8 characters with �

eed3si9n (approved):
Thanks @wrobstory!

crates/python_extractor/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+            expand_path_to_defs("/Users/foo/bar/src/main/python/blah/ppp.py"),
+            expected
+        );
+    }
+
+    #[test]
+    fn expand_path_to_defs_ambigious_path_test() {
+        let mut expected = vec!["src.main.python.blah.src.main.ppp", "src.main.ppp"];
+        expected.sort();
+        expected.dedup();
```
johnynek, line 164: why do we need this? That list doesn't have duplicates.

crates/python_extractor/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+            expand_path_to_defs("/Users/foo/bar/src/main/python/blah/src/main/ppp.py"),
+            expected
+        );
+    }
+
+    #[test]
+    fn expand_site_packages_path_to_defs_test() {
+        let mut expected = vec!["pytz", "pytz.__init__"];
+        expected.sort();
+        expected.dedup();
```
johnynek, line 176: same comment.

crates/python_extractor/tests/data/nonascii.py
```diff
@@ -0,0 +1,4 @@
+# coding: iso-8859-5
```
johnynek, line 1: could we not detect this and use the correct character set to decode?

https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations


> If a comment in the first or second line of the Python script matches the regular expression coding[=:]\s*([-\w.]+),  this comment is processed as an encoding declaration; the first group of this expression names the encoding of the source code file.


Seems easy enough to parse with a regex. Take the first two lines and look for `#.*coding[=:]\s*([-\w.]+)` and then look up the character set by string. If it isn't found, then use utf-lossy like you have here.

johnynek (approved):
We can ship if it unblocks, but I would like to be a bit more pro.

crates/python_extractor/src/lib.rs
wrobstory: this was written by someone else, but i will fix it 😉

crates/python_extractor/src/lib.rs
```diff
@@ -1,18 +1,74 @@
...
-use bzl_gen_build_python_utilities::PythonProgram;
-
-use bzl_gen_build_shared_types::api::extracted_data::{DataBlock, ExtractedData};
-
-use anyhow::{Context, Result};
-
 mod extract_py_bzl_gen_build_commands;
 mod extract_py_imports;
 
+fn get_codec_map() -> HashMap<&'static str, &'static Encoding> {
```
johnynek, line 14: can this be `const`?

I asked gpt and it said something like this would work:
```rust
use std::collections::HashMap;
use lazy_static::lazy_static;

lazy_static! {
    static ref MY_MAP: HashMap<&'static str, &'static str> = {
        let mut m = HashMap::new();
        m.insert("key1", "value1");
        m.insert("key2", "value2");
        m
    };  
}

```

crates/python_extractor/src/lib.rs
```diff
@@ -44,20 +101,47 @@ pub async fn extract_python(
...
+        reader.read_line(&mut first_line).with_context(|| {
+            format!(
+                "While attempting to read first line from file to check for encoding: {:?
+            }",
                 input_file
             )
         })?;
 
-        let input_str = String::from_utf8_lossy(&buffer).into_owned();
+        let encoding = if first_line.starts_with("# coding:") {
```
johnynek, line 128: seems like we went pretty far here, what about using the regex from the standard something like `.*#.*\bcoding[=:]\s*([-\w.]+)` using the regex crate?

johnynek (approved):
Looks good. Thanks for adding the table. 

I added one more suggestion if you want to take, but if you're sick of it, then merge and move on.

johnynek (approved):
This looks great!

Thanks for going the extra mile.

Issue #169, opened 2023-12-06
Support polyglot BUILD.bazel ("Google3" layout)


eed3si9n:
## problem

In a "Google3" layout, the source files are supposedly organized by their package names without language tags `scala/` or distinction of `main` vs `test`:

- `com/example/`
   - `foo/`
        - `a.proto`
        - `b.py`
        - `C.scala`

Currently bzl-gen-build does not support Google3-like layout because one run would overwrite the `BUILD.bazel` file.

## proposal

I wonder if we can support this use case if we implemented `--append` flag, and keep appending targets on the BUILD file.

eed3si9n:
Currently the Google3 layout to colocate "main" and test targets in one package is limited to `--no-aggregate-source` mode. This works ok for cases like Protobuf, but I think we also want to aggregate by package in languages like Scala and Java, and potentially even Python to reduce the risk of dangling source files (sources that are added to git but not tracked by any target).

PR #170, opened 2023-12-08
Support polyglot BUILD file

eed3si9n:
Fixes https://github.com/bazeltools/bzl-gen-build/issues/169

**Problem**
Currently it's not possible to mix languages within a package.

**Solution**
1. This implements `--append` flag to allow appending contents to BUILD.bazel.
2. This also adds `include_file_extension` configuration to allow `py_test` target to keep the name of the file as the target name.


crates/driver/src/print_build.rs
```diff
@@ -259,48 +259,51 @@ where
...
-        let mut extra_kv_pairs: HashMap<String, Vec<String>> = HashMap::default();
+            to_name_from_file_name(&node_file_name, include_file_extension)?
+        }; 
 
         fn add_non_empty(
             opt: &'static Opt,
             key: &str,
             labels: &Vec<String>,
             extra_kv_pairs: &mut HashMap<String, Vec<String>>,
+            include_file_extension: bool,
```
johnynek, line 290: could we avoid boolean blindness with a small Enum to represent the semantics:
```rust
enum TargetNameStrategy { Directory, DirectoryFileType }
```
or something

crates/driver/src/print_build.rs
```diff
@@ -510,12 +513,23 @@ where
             } else {
                 let directory = to_directory(entry.to_string());
                 let full_file = opt.working_directory.join(&entry);
-                let name = to_file_name(&full_file).replace(".", "_");
+                let name = to_name_from_file_name(&to_file_name(&full_file), include_file_extension).unwrap();
```
johnynek, line 516: can we remove the unwrap or comment why it will never happen?

crates/driver/src/print_build.rs
```diff
@@ -853,13 +885,14 @@ mod tests {
     use crate::Commands::PrintBuild;
     use std::collections::BTreeMap;
 
-    fn example_opt(no_aggregate_source: bool) -> Opt {
+    fn example_opt(no_aggregate_source: bool, append: bool) -> Opt {
```
johnynek, line 888: similar to the boolean blindness.. What about an `enum WriteMode { Overwrite, Append }`

crates/python_extractor/src/lib.rs
```diff
@@ -220,6 +220,9 @@ fn expand_path_to_defs(path: &str) -> Vec<String> {
         if element == "src" {
             results.push("src".to_string());
         }
+        if element == "com" {
```
johnynek, line 223: can we comment about this special case. I don't understand it.

crates/driver/src/print_build.rs
```diff
@@ -510,12 +513,23 @@ where
...
             }
         }
     }
 
+    fn to_name_from_file_name(file_name: &String, include_file_extension: bool) -> Result<String> {
+        if include_file_extension {
+            Ok(file_name.replace(".", "_"))
+        } else {
+            match Path::new(&file_name).file_stem() {
+                Some(s) => Ok(s.to_str().unwrap().to_string()),
```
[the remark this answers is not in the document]
eed3si9n: At least I'm hoping that's the case.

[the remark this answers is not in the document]
eed3si9n: Addressed in d5f87f4d9efa5e17aeb85d1e1e220ef650c86d4d

crates/driver/src/print_build.rs
eed3si9n: It's highly unlikely but I added `unwrap_or()` and also commented in d5f87f4d9efa5e17aeb85d1e1e220ef650c86d4d

crates/driver/src/print_build.rs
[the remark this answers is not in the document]
eed3si9n: `s.to_str()` returns an `Option`, and my guess is that it's highly unlikely that it would fail.

crates/python_extractor/src/lib.rs
eed3si9n: Fixed the hardcoding in 65e003cfa96d0959e709d1bacd37c944e4f47199. To turn `/Users/foo/bar/src/main/python/blah/aaa.py` into the Python package `src.main.python.blah.ppp` previously it was looking for the directory name `src/`. I've fixed it so the function accepts relative path instead to know where the cutoff should be.

example/com/example/BUILD.bazel
```diff
@@ -5,3 +5,6 @@ proto_library(name='aa_proto', srcs=['aa.proto'], visibility=['//visibility:publ
 java_proto_library(name='aa_proto_java', deps=[':aa_proto'], visibility=['//visibility:public'])
 py_proto_library(name='aa_proto_py', deps=[':aa_proto'], visibility=['//visibility:public'])
 # ---- END BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
+# ---- BEGIN BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
+py_library(name='hello', srcs=['hello.py'], deps=['@@rules_python~0.24.0~pip~pip_39_pandas//:pkg'], visibility=['//visibility:public'])
+# ---- END BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
```
non, line 10: Is it expected to have two different generated sections with the same name?

I had thought we'd only have one of these, or else we'd use two different names for the different parts. (That way each generated section could be safely updated independently without affecting any others.)

non (commented):
The changes themselves seemed OK but I had a question about the expected behavior/usage of the tool.

example/com/example/BUILD.bazel
eed3si9n: No the target name must be unique within a package. This package demonstrates that bzl-gen-build can be configured to produce "hello" package even though it's inside of the "example" package. Normally we make a target based on the package (directory) name not the source file name.

example/build_tools/lang_support/create_lang_build_files/regenerate_python_build_files.sh
```diff
@@ -32,4 +32,6 @@ bazel query '@pip//...' | grep '@pip' > $TMP_WORKING_STATE/external_targets
     # update_remote_cache "remote_python_${CACHE_KEY}"
 # fi
 
-run_system_apps "build_tools/lang_support/create_lang_build_files/bazel_${GEN_FLAVOR}_modules.json"
+run_system_apps "build_tools/lang_support/create_lang_build_files/bazel_${GEN_FLAVOR}_modules.json" \
+  --no-aggregate-source \
```
eed3si9n, line 36: @non This is the option.

example/com/example/BUILD.bazel
non: Sorry i wasn't more clear. I was actually referring to the generated comment section names (i.e. "BZL_GEN_BUILD_GENERATED_CODE"). If we have to generators "sharing" a file and want them to safely be able to update their own section they'd need to be able to distinguish those sections somehow.

It's not really a big deal,  and it would be easy to change I think, I just wanted to point it out.

crates/driver/src/print_build.rs
```diff
@@ -259,47 +263,60 @@ where
...
+        ) -> Result<()> {
             if !labels.is_empty() {
-                let vals = labels.iter().map(|e| to_label(&opt, e)).collect();
+                let vals = labels
+                    .iter()
+                    .map(|e| to_label(&opt, e, target_name_strategy))
+                    .collect();
                 extra_kv_pairs.insert(key.to_string(), vals);
             }
+            Ok(())
```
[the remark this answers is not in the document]
eed3si9n: Nice catch. I think at some point I tried to make this function return `Result<()>`, but it no longer contains a `?` anymore so we can probably remove `Ok(())`.

example/com/example/BUILD.bazel
eed3si9n: At least in the way it's done right now, the polyglot support is implemented by simply appending output after another, so yea, we end up with multiple codegen regions. I guess it's something we can follow up.

crates/driver/src/print_build.rs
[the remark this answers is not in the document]
eed3si9n: Addressed in 64784de.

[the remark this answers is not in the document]
non: Good catch!

non (approved):
Looks OK to me (once main is merged). I'd like it if the polyglot appenders used unique names but that's easy to do later.

wrobstory (approved):

PR #174, opened 2023-12-20
Readme improvement

eed3si9n:
This adds some documentation about how to configure bzl-gen-build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,39 +1,57 @@
-# Modular build generator
+bzl-gen-build
+=============
 
-This is a modular build generator, it is designed so it is easy to pick and choose components, modify intermediate states (all json), and post-process or use the outputs for new purposes as it makes sense.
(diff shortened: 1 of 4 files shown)


non (approved):
This looks reasonable to me.

It might have been nice to split the change up between documentation and functionality. From what I can tell the script changes are about clearing and then re-downloading the external files, which we want to do when regenerating the builds. Is that right? Am I missing something?

eed3si9n:
Re: script changes

1. The script references `fetch_binary()` but was not defined, since we normally just the freshly built on instead of downing, so I added an implementation of `fetch_binary()` here.
2. The other change about `CACHE_KEY` is similar story. A commented out code references `CACHE_KEY` but it wasn't defined, so I added it to demonstrate what should be used as the cache key, if ppl want to cache some JSON files.

PR #188, opened 2024-01-03
Support main and test within BUILD.bazel ("Google3")

eed3si9n:
Problem
-------
In a "Google3" layout, the source files are supposedly organized by their package names, regardless of main or test.

Solution
--------
This implements a new configuration called `test_globs` so both main and test can point at a same directory tree when running with `--no-aggregate-source`. This allows "Google3" layout for both Protobuf and Python. See `example/`.


crates/driver/src/extract_defrefs.rs
```diff
@@ -580,3 +619,71 @@ pub async fn extract_defrefs(
...
+        on_entry: F,
+    ) -> Result<Vec<A>>
+    where
+        F: Fn(DirEntry, PathBuf) -> R,
+        R: futures::Future<Output = Result<A>> + Send + 'static,
+    {
+        let mut results: Vec<A> = Vec::default();
+        let globset = to_globset(test_globs)?;
+        let is_empty_globs = test_globs.is_empty();
+        for entry in WalkBuilder::new(working_directory.join(child_path))
```
johnynek, line 643: can we remove the duplication of this WalkBuilder filter? It's the same as the one above isn't it?

example/com/example/BUILD.bazel
```diff
@@ -8,3 +8,6 @@ py_proto_library(name='aa_proto_py', deps=[':aa_proto'], visibility=['//visibili
 # ---- BEGIN BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
 py_library(name='hello', srcs=['hello.py'], deps=['@@rules_python~0.24.0~pip~pip_39_pandas//:pkg'], visibility=['//visibility:public'])
 # ---- END BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
+# ---- BEGIN BZL_GEN_BUILD_GENERATED_CODE ---- no_hash
+py_test(name='hello_test', srcs=['hello_test.py'], deps=['//com/example:hello'], visibility=['//visibility:public'])
```
eed3si9n, line 12: Test co-existing under `com/example/` package.

crates/driver/src/print_build.rs
```diff
@@ -236,36 +238,39 @@ where
...
-    let (module_config, path_type, _matched_prefix) = if let Some(a) = module_config {
+    let module_config = if let Some(a) = module_config {
         a
     } else {
-        return Err(anyhow!(
-            "Unable to find any matching configuration for {}",
-            element
-        ));
+        let t: TargetEntries = Default::default();
+        return Ok(t);
```
johnynek, line 256: nit: `return Ok(Default::default());` or maybe even `return Default::default();` would work?

crates/driver/src/print_build.rs
```diff
@@ -435,7 +439,14 @@ where
                         };
                         t.entries.push(filegroup_target);
                     } else {
-                        parent_include_src.push(format!("{}", node_file_name));
+                        if is_empty_globs
```
johnynek, line 442: this is also duplicated, I wonder if this expression could be a function on something instead.

crates/driver/src/print_build.rs
```diff
@@ -766,8 +777,25 @@ async fn print_file(
     let t = generate_targets(
         opt,
         project_conf,
-        graph_nodes,
-        element,
+        SourceConfig::Main,
+        graph_nodes.clone(),
+        element.clone(),
```
johnynek, line 782: if you clone element every time, might as well take a `&String` or `&str` instead IMO.

crates/driver/src/extract_defrefs.rs
```diff
@@ -141,43 +142,75 @@ async fn process_file(
     }
 }
 
+// check that a file entry is a match
+fn has_good_extension(entry: &DirEntry, file_extensions: &Vec<OsString>) -> bool {
+    if entry.file_type().map(|e| e.is_file()).unwrap_or(false) {
```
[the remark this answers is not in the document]
eed3si9n: Addressed in f6206768de7b16fc1630c0e657597e07f2474d91

crates/driver/src/extract_defrefs.rs
```diff
@@ -141,43 +142,75 @@ async fn process_file(
...
+        false
+    }
+}
+
+pub fn to_globset(test_globs: &Vec<String>) -> Result<GlobSet> {
+    if test_globs.is_empty() {
+        let globset_builder = &mut GlobSetBuilder::new();
+        Ok(globset_builder.add(Glob::new("**/*.*")?).build()?)
+    } else {
+        let globset_builder = &mut GlobSetBuilder::new();
```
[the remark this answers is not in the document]
eed3si9n: Addressed in 8a28a9491c8051b5f3247173d303ff03be0cb466

crates/driver/src/print_build.rs
eed3si9n: Split up into a function in 9a9ed9cecef933e111d92232c11b257c095e1cc5

crates/driver/src/extract_defrefs.rs
```diff
@@ -141,43 +142,75 @@ async fn process_file(
...
     > = Vec::default();
+    let is_empty_globs = opt.module_config.test_globs.is_empty();
+    let globset = to_globset(&opt.module_config.test_globs)?;
     for entry in WalkBuilder::new(working_directory.join(child_path))
         .build()
         .into_iter()
         .filter_map(|e| e.ok())
+        .filter(|e| has_good_extension(e, &opt.file_extensions))
+        .filter(|e| {
+            is_empty_globs || globset.is_match(e.path()) ^ (source_config == SourceConfig::Main)
```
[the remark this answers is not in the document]
eed3si9n: Addressed in 9a9ed9cecef933e111d92232c11b257c095e1cc5

crates/driver/src/print_build.rs
eed3si9n: Addressed in 9a9ed9cecef933e111d92232c11b257c095e1cc5

crates/driver/src/print_build.rs
eed3si9n: Addressed in ab7edd93aaef9c1b50592e662193c0cc0dfa135c

crates/driver/src/extract_defrefs.rs
eed3si9n: Yea, so there's an identical block of code in `async fn async_extract_def_refs` and `walk_directories` (in test). I did originally implement `walk_directories` intending to split out and write tests for glob evaluation logic, but ran into ownership issue of accessing `opt: Arc<ExtractConfig>` from a lambda, which kept complaining that the lambda would outlive `opt`, so I've split out `has_good_extension` and `to_globset` so at least we have some tests. That was last year.

I can try again to convince the compiler that `opt` would live longer. Each time I try to make things testable by making function accept a lambda I run into this. Maybe I did `opt: &'static Opt,` last time?

johnynek (approved):

crates/driver/src/extract_defrefs.rs
eed3si9n: Addressed in dbf1e05.

PR #204, opened 2024-01-19
Minor cleanups

johnynek:
Two kinds of patterns I cleaned up:

1. remove `spawn(async(` when possible. Sometimes it isn't possible and I added a comment.
2. remove unneeded `clone` calls.

### DIFF
--- a/crates/driver/src/extract_defrefs.rs
+++ b/crates/driver/src/extract_defrefs.rs
@@ -96,15 +96,13 @@ async fn process_file(
     };
     let st = Instant::now();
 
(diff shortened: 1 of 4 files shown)


crates/driver/src/extract_defrefs.rs
```diff
@@ -624,20 +617,15 @@ pub async fn extract_defrefs(
         }
 
         for (entry, files) in work.into_iter() {
-            let sha_of_conf_config = sha_of_conf_config.clone();
-            let path_sha_to_merged_defrefs = path_sha_to_merged_defrefs;
```
non, line 628: 😅

non (approved):

PR #209, opened 2024-01-30
Optionally parallelize scala and java parsing

johnynek:
in our monorepo, the parsing of these files is the lion's share of the code. This introduces an option to load all the string into memory, then in parallel, parse all the strings. Since the parsing is non-blocking, cats-effect should use the cpu thread-pool and not do anything crazy.

If you run into problems you can add `--sequential` to get the old behavior.

### DIFF
--- a/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
+++ b/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
@@ -1,7 +1,7 @@
 package io.bazeltools.buildgen.scaladefref
 
-import cats.Monad
-import cats.data.{Chain, NonEmptyList, State, Writer}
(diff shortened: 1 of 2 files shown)


language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/scaladefref/ScalaSourceEntityExtractor.scala
```diff
@@ -23,33 +23,83 @@ import cats.syntax.all._
...
+        (allMessages(first) ++ allMessages(second))
+          .mkString_("combined ScalaSourceEntityExtractor errs:", "\n\t", "\n")
+      )
+
+  implicit val semigroupErr: Semigroup[Err] =
+    new Semigroup[Err] {
+      def combine(left: Err, right: Err): Err = CombinedErr(left, right)
+    }
+
+  def parseDirectives(src: Source): IO[Chain[String]] = {
```
johnynek, line 62: this change is in order to remove a `sys.error`

wrobstory (approved):

PR #218, opened 2024-02-14
relative imports test demo

vors:
Hello!

This is a quick repro for existing problem with relative imports: they are treated like non-relative imports at the moment, but ideally they should take into account the passed filename.

I think it's a limitation of used python parser library from what i can tell.


vors:
@ianoc mentioned he may take a look when he has time

PR #219, opened 2024-02-17
Add tests for new python syntax

vors:
Hello!

Here are tests that are legal python syntax (i.e. https://github.com/astral-sh/ruff parses it fine) but they are currently failing

```
thread 'extract_py_imports::tests::test_syntax_constructs' panicked at python_extractor/src/extract_py_imports.rs:162:68:
called `Result::unwrap()` on an `Err` value: invalid syntax. Got unexpected token '[' at line 7 column 30
```

I'm not sure that currently used parser is a good long-term choice to avoid that kind of problems.
Maybe we can switch to ruff parser altogether?

### DIFF
--- a/crates/python_extractor/src/extract_py_imports.rs
+++ b/crates/python_extractor/src/extract_py_imports.rs
@@ -145,4 +145,27 @@ def my_fn():
         expected.dedup();
         assert_eq!(extract(&parsed), expected)
     }
+
+    #[test]
+    fn test_syntax_constructs() {
+        let python_source = r#"
+from typing import Union
+from foo import perform_cast
+value = 1
+args = [int, float]
+# this is a legal-in-runtime construction
+def perform_cast(value, Union[*args])
+# walrus operator
+x := y := 1
+        "#;
+
+        let parsed = PythonProgram::parse(python_source, "tmp.py").unwrap();
+        let mut expected = vec![
+            "typing.Union".to_string(),
+            "foo.perform_cast".to_string(),
+        ];
+        expected.sort();
+        expected.dedup();
+        assert_eq!(extract(&parsed), expected)
+    }
 }


eed3si9n:
I think we use RustPython's parser, but the dependency is pinned to this commit - https://github.com/RustPython/RustPython/commit/b9ed63ed326e4ab9c97d808271ddc1d7ca05fda7, which is 1409 commits behind https://github.com/RustPython/RustPython/compare/b9ed63ed326e4ab9c97d808271ddc1d7ca05fda7...main.

johnynek:
it looks like ruff doesn't seem to have an external parser dependency:

https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_parser/Cargo.toml

I wonder if we can depend on their parser since I think that is fairly well tested across a bunch of code bases.

vors:
Another alternative would be to depend on the native CPython ast parser and dump it into json with something like https://pypi.org/project/ast2json/ then parse import in rust or perhaps just do this whole thing in python, but that would mean starting brining dependencies outside of rust ecosystem which is big design decision.

crates/python_extractor/src/extract_py_imports.rs
```diff
@@ -145,4 +145,27 @@ def my_fn():
...
+
+    #[test]
+    fn test_syntax_constructs() {
+        let python_source = r#"
+from typing import Union
+from foo import perform_cast
+value = 1
+args = [int, float]
+# this is a legal-in-runtime construction
+def perform_cast(value, Union[*args])
```
johnynek, line 157: this is legal?

johnynek:
I merged main to pickup: https://github.com/bazeltools/bzl-gen-build/pull/237

That should fix match issues, but your test surprises me that it is legal syntax.

vors:
Thank you, yes the `[*foo]` is legal syntax but pyright for example also has troubles with it at the moment. Ruff doesn't (and python 3.11.8 doesn't)

PR #227, opened 2024-02-28
Improve Python indexing example

eed3si9n:
## Problem
In the current example Python indexing only tracks 3rdparty libraries, but not Protobuf-generated sources.

## Solution
This adjusts the scripts to index both 3rdparty pip targets and Protobuf-generated sources.

### DIFF
--- a/example/build_tools/bazel_rules/wheel_scanner/wheel_scanner.bzl
+++ b/example/build_tools/bazel_rules/wheel_scanner/wheel_scanner.bzl
@@ -1,7 +1,4 @@
 def _wheel_scanner_impl(target, ctx):
-    if not target.label.workspace_name.startswith("rules_python"):
-        return []
-
     # Make sure the rule has a srcs attribute.
     out_content = ctx.actions.declare_file("%s_wheel_scanner.json" % (target.label.name))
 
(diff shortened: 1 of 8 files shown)


example/build_tools/bazel_rules/wheel_scanner/wheel_scanner.bzl
```diff
@@ -21,6 +18,14 @@ def _wheel_scanner_impl(target, ctx):
                     fail("Didn't have workspace prefix")
                 all_py_relative_paths.append(path[len(workspace_root) + 1:])
                 all_py_files.append(file)
+    elif ctx.rule.kind == "py_proto_library":
```
eed3si9n, line 21: The main changes are here. Instead of scanning for Python files in wheel, I'm grabbing the first file in DefaultInfo of the target.

example/com/example/hello.py
```diff
@@ -1,4 +1,5 @@
 import pandas as pd
+from com.example.aa_pb2 import A
```
eed3si9n, line 2: Here's a demo usage.

example/build_tools/bazel_rules/wheel_scanner/wheel_scanner.bzl
johnynek: why not add all files and let the command sort it out?

johnynek (approved):

example/build_tools/bazel_rules/wheel_scanner/wheel_scanner.bzl
eed3si9n: `DefaultInfo#files` returns all Python files including from the dependency, so import would end up adding a bunch of unnecessary deps.

PR #254, opened 2024-04-30
Small update to "We cannot find parent dir" error message

ramyanaga:
Problem:
I had some trouble debugging an oncall question related to this issue, and couldn't have figured it out without someone telling me when this error usually occurs.

Solution:
This PR adds some more detail to the error message


johnynek:
Looks like the CI needs to be updated for some jdk 8 issues.

johnynek (approved):

PR #264, opened 2024-05-09
Refactor python extraction a bit for smaller functions

johnynek:


crates/driver/src/print_build.rs
```diff
@@ -610,11 +615,14 @@ where
...
-                if let Some(ep) = &bin.binary_refs.target_value {
-                    k_strs.push(("binary_refs_value".to_string(), ep.to_string()));
+                if let Some(tv) = &bin.binary_refs.target_value {
+                    k_strs.push(("binary_refs_value".to_string(), tv.to_string()));
                 }
 
-                k_strs.push(("owning_library".to_string(), format!(":{}", lib_target.to_string())));
+                k_strs.push((
+                    "owning_library".to_string(),
+                    format!(":{}", lib_target),
```
johnynek, line 624: don't need `lib_target.to_string()` here.

crates/driver/src/print_build.rs
```diff
@@ -610,11 +615,14 @@ where
                     k_strs.push(("entity_path".to_string(), ep.to_string()));
                 }
 
-                if let Some(ep) = &bin.binary_refs.target_value {
-                    k_strs.push(("binary_refs_value".to_string(), ep.to_string()));
+                if let Some(tv) = &bin.binary_refs.target_value {
```
johnynek, line 618: we copied `ep` but that seems to come from entity_path above.

crates/python_extractor/src/lib.rs
```diff
@@ -71,7 +72,7 @@ lazy_static! {
     };
 }
 
-pub fn read_file_to_str(input_file: &PathBuf) -> Result<String, anyhow::Error> {
+pub fn read_file_to_str(input_file: &Path) -> Result<String, anyhow::Error> {
```
johnynek, line 75: `&PathBuf` is an anti-pattern, use `&Path` which is a reference to an immutable Path.

crates/python_extractor/src/lib.rs
```diff
@@ -120,68 +121,90 @@ pub fn read_file_to_str(input_file: &PathBuf) -> Result<String, anyhow::Error> {
     }
 }
 
-pub async fn extract_python(
-    relative_input_paths: String,
-    working_directory: PathBuf,
-    output: PathBuf,
-    label_or_repo_path: String,
+async fn extract_file(
```
johnynek, line 124: write a function that just runs on a single file to reduce the size of the function below that calls it.

eed3si9n (approved):

crates/python_extractor/src/lib.rs
```diff
@@ -120,68 +121,90 @@ pub fn read_file_to_str(input_file: &PathBuf) -> Result<String, anyhow::Error> {
...
-        } else {
-            defs.extend(expand_path_to_defs(&relative_path, file_p.as_ref()));
-        }
-        data_blocks.push(DataBlock {
-            entity_path: relative_path,
-            defs,
-            refs,
-            bzl_gen_build_commands,
-        })
+    for data_block in join_all(all_loads).await {
```
johnynek, line 202: use join_all to run extraction in parallel rather than sequentially.

PR #319, opened 2024-09-23
Sort kwargs like buildifier

johnynek:
buildifier, for some reason, puts `runtime_deps` before `deps`.

In order to print things in an order such that buildifier wouldn't change them, we need to do the same.

I found this sorting table in the buildifier code and just used the same order.

### DIFF
--- a/crates/driver/src/print_build.rs
+++ b/crates/driver/src/print_build.rs
@@ -42,6 +42,60 @@ struct TargetEntry {
 }
 
 impl TargetEntry {
+    fn sort_like_buildifier<E>(kwargs: &mut Vec<(Arc<String>, E)>) {
+        // this is from: https://github.com/bazelbuild/buildtools/blob/be1c24cc9a44b4fd2410ec5356e4e21926dd206a/tables/tables.go#L177C34-L210C2
+        fn priority(key: &str) -> i32 {
+            match key {
+                "name" => -99,
+                "archive_override.module_name" => -99,
+                "git_override.module_name" => -99,
+                "local_path_override.module_name" => -99,
+                "multiple_version_override.module_name" => -99,
+                "single_version_override.module_name" => -99,
+                "bazel_dep.version" => -98,
+                "module.version" => -98,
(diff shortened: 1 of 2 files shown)


eed3si9n (approved):

crates/driver/src/print_build.rs
```diff
@@ -668,11 +723,11 @@ where
 
     fn append_key_values(
         extra_kv_pairs: &mut HashMap<String, Vec<String>>,
-        key: &String,
+        key: String,
         values: &Vec<String>,
     ) {
         extra_kv_pairs
-            .entry(key.clone())
```
johnynek, line 675: if we unconditionally clone, just accept the `String` so the caller can possibly avoid the clone.

crates/driver/src/print_build.rs
```diff
@@ -786,10 +841,11 @@ where
 async fn print_file(
     opt: &'static Opt,
     project_conf: &'static ProjectConf,
-    graph_nodes: Vec<GraphNode>,
+    mut graph_nodes: Vec<GraphNode>,
     concurrent_io_operations: &'static Semaphore,
     element: String,
 ) -> Result<Vec<PathBuf>> {
+    graph_nodes.sort_by(|a, b| a.node_label.cmp(&b.node_label));
```
johnynek, line 848: sorting is moved here rather than after adding each item into `graph_nodes`

crates/driver/src/print_build.rs
```diff
@@ -927,15 +983,13 @@ pub async fn print_build(
...
     {
-        let graph_node = graph_node.clone();
         let element = if !opt.no_aggregate_source {
             entry
         } else {
             to_directory(entry.to_string())
         };
         let v = graph_nodes.entry(element).or_default();
         v.push(graph_node);
-        v.sort_by(|a, b| a.node_label.cmp(&b.node_label))
```
johnynek, line 938: this is moved inside the printing function. It isn't great to call sort after adding each element. We should only do it once.

PR #321, opened 2024-09-23
Sort labels like buildifier does

johnynek:
buildifier sorts labels by package, then the rule, not by treating the label like a string and sorting that.

This improves the sorting and normalization of labels to avoid `foo:foo` situations.

### DIFF
--- a/crates/python_utilities/src/lib.rs
+++ b/crates/python_utilities/src/lib.rs
@@ -3,10 +3,17 @@ use rustpython_parser::{ast, Parse};
 
 pub mod ast_builder;
 pub mod ast_printer;
+use ast_printer::{WritingBuffer, emit_body};
 
-#[derive(Debug, PartialEq)]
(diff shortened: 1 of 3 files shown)


crates/python_utilities/src/lib.rs
```diff
@@ -3,15 +3,28 @@ use rustpython_parser::{ast, Parse};
...
 pub mod ast_printer;
+use ast_printer::{WritingBuffer, emit_body};
 
-#[derive(Debug, PartialEq)]
+#[derive(PartialEq)]
 pub struct PythonProgram {
     pub body: Vec<Stmt>,
 }
 
+impl std::fmt::Display for PythonProgram {
```
johnynek, line 13: this implementation should be with the type, not in another file.

johnynek:
sorry for merging this without review. I *assumed* that it would block on one PR approval.

PR #322, opened 2024-09-24
Don't unconditionally add a new line in build generation when we don't load

johnynek:


crates/python_utilities/src/ast_printer.rs
```diff
@@ -338,6 +349,18 @@ filegroup(
         )
     }
 
+    #[test]
+    fn round_trip_build_file_fg() {
```
johnynek, line 353: this test fails on main.

eed3si9n (approved):

PR #376, opened 2025-02-13
Assume "com" imports are never relative.

non:
Previously, ever import that occurs after a wildcard import was tested to see if it was a relative import, such as:

    import com.acme._
    import ip.tools.Device // imports com.acme.ip.tools.Device

This causes problems when there are multiple wildcard imports folowed by a lot of imports, since you generate a combinatorial explosion of potential entities:

    import alpha._
    import beta._
    import gamma._

	import com.acme.ip.tools.Device
	// potential entities from this one import:
	// - com.acme.ip.tools.Device
	// - alpha.com.acme.ip.tools.Device
	// - beta.com.acme.ip.tools.Device
	// - gamma.com.acme.ip.tools.Device
	// - alpha.beta.com.acme.ip.tools.Device
	// - alpha.gamma.com.acme.ip.tools.Device
	// - beta.gamma.com.acme.ip.tools.Device
	// - alpha.beta.gamma.com.acme.ip.tools.Device

It is very rare to see "com" appear anywhere other than as the TLD, and it is even more rare to see an import that starts with com where it is not used as a TLD. So as a heuristic, this PR introduces a change where specially for imports starting with "com" we will not consider any previous wildcard imports.

**EDIT: Rather than hardcoding `com` we are now supporting configuration using the environment variable `BZL_GEN_SPECIAL_TLDS` to support zero-or-more of these.**

We still support explicit imports such as:

    import com.acme.net.com.Device

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,6 +194,8 @@ As an optimization, setting the environment variable `BZL_GEN_SPECIAL_TLDS` to a
 export BZL_GEN_SPECIAL_TLDS=com,net,org
 ```
 
+The names must match the regular expression `^[a-z]+$`.
+
 ### Extracting definitions from 3rdparty libraries
 
 Extracting definitions from 3rdparty libraries require some setup, also demonstrated in the `example/` repo.
--- a/README.md
+++ b/README.md
@@ -184,6 +184,16 @@ In some situations, like for Protocol Buffer schemas, we want to generate second
 }
 ```
 
+### Heuristics
+
+Wildcard imports in Java and Scala can be expensive to resolve, since every subsequent import migth be relative to the previous wildcard.
+
+As an optimization, setting the environment variable `BZL_GEN_SPECIAL_TLDS` to a comma-separated list will tell the driver program to assume any import rooted at one of those names is not relative. For example:
+
+```bash
+export BZL_GEN_SPECIAL_TLDS=com,net,org
+```
+
 ### Extracting definitions from 3rdparty libraries
 
 Extracting definitions from 3rdparty libraries require some setup, also demonstrated in the `example/` repo.
--- a/README.md
+++ b/README.md
@@ -186,7 +186,7 @@ In some situations, like for Protocol Buffer schemas, we want to generate second
 
 ### Heuristics
 
-Wildcard imports in Java and Scala can be expensive to resolve, since every subsequent import migth be relative to the previous wildcard.
+Wildcard imports in Java and Scala can be expensive to resolve, since every subsequent import might be relative to the previous wildcard.
 
 As an optimization, setting the environment variable `BZL_GEN_SPECIAL_TLDS` to a comma-separated list will tell the driver program to assume any import rooted at one of those names is not relative. For example:
 
--- a/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DriverApplication.scala
+++ b/language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DriverApplication.scala
@@ -13,7 +13,7 @@ import io.circe.syntax.EncoderOps
 
 abstract class DriverApplication extends IOApp {
   def name: String
-  def extract(data: String): IO[Symbols]
+  def extract(data: String, specialTlds: Set[String]): IO[Symbols]
 
   def readToString(path: Path): IO[String] =
     IO.blocking(
@@ -30,7 +30,8 @@ abstract class DriverApplication extends IOApp {
 
   private[this] def parallelExtractDataBlocks(
       workingDirectory: Path,
-      paths: List[String]
+      paths: List[String],
+      specialTlds: Set[String]
   ): IO[List[DataBlock]] =
     paths.sorted
       .traverse { path =>
@@ -46,7 +47,7 @@ abstract class DriverApplication extends IOApp {
         // which is limited to the number of CPUs
         // we can parse all the code
         inMemory.parTraverse { case (path, content) =>
-          extract(content).attempt
+          extract(content, specialTlds).attempt
             .flatMap {
               case Right(x) => IO.pure(x)
               case Left(err) =>
@@ -63,14 +64,15 @@ abstract class DriverApplication extends IOApp {
 
   private[this] def sequentialExtractDataBlocks(
       workingDirectory: Path,
-      paths: List[String]
+      paths: List[String],
+      specialTlds: Set[String]
   ): IO[List[DataBlock]] =
     paths.sorted
       .traverse { path =>
         val fullPath = workingDirectory.resolve(path)
         for {
           content <- readToString(fullPath)
-          try_e <- extract(content).attempt
+          try_e <- extract(content, specialTlds).attempt
           symbols <- try_e match {
             case Right(x) => IO.pure(x)
             case Left(err) =>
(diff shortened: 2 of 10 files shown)


language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/Entity.scala
```diff
@@ -37,6 +37,33 @@ final case class Entity(parts: Vector[String]) {
...
+  // Note that we must avoid breaking imports like:
+  //
+  //    import com.foo.com.bar.Qux
+  //    import com.acme.shadow.com.google.Dingus
+  //
+  // We could also handle other common TLDs such as "net" and "org"
+  // the same way but "com" is the most common and one of the least
+  // likely to occur as a "split import".
+  final val SpecialCom = "com"
+  final val SpecialComEntity = Entity.Resolved.Known(Entity.simple("com"))
```
jklukas, line 65: The code commentary here is great and explains what's going on. I worry, though, that future users won't have any idea this exists to go looking for it, and the behavior is pretty subtle.

Should this behavior be controlled by a configuration option and be off by default? Perhaps the ideal would be that we have an option called something like "non_wildcard_prefix_hints" that would take an array like `["com"]`.

I do take some comfort that the README mentions "Scala/Java can have a lot of references as they are often heuristic-based when we have limited insights (wildcard imports)". I think it's reasonable to move forward with the change as written here if we add a line to the documentation that some of the wildcard import heuristics are lossy, so we aren't guaranteed to pick up all wildcard imports.

non: Yeah it's a good point. I'm not sure what configuration options we have or how hard it would be to plumb them through to the code, but I will investigate further.

language_generators/scala-defref-extractor/src/test/scala/io/bazeltools/buildgen/scaladefref/CanParseFileTest.scala
```diff
@@ -58,11 +58,7 @@ class CanParseFileTest extends AnyFunSuite {
         Entity.dotted("com"),
         Entity.dotted("com.example"),
         Entity.dotted("com.example.Elephant"),
-        Entity.dotted("com.example.Wolf"),
-        Entity.dotted("com.foo.bar.com"),
-        Entity.dotted("com.foo.bar.com.example"),
-        Entity.dotted("com.foo.bar.com.example.Elephant"),
-        Entity.dotted("com.foo.bar.com.example.Wolf")
+        Entity.dotted("com.example.Wolf")
```
jklukas, line 61: This is fun to see that no new tests are necessary, but the change naturally cleans up a bunch of undesirable lines in the existing tests.

jklukas (approved):

johnynek:
Just a comment:

Another design would be to have a set of "known roots" and for instance "_root_" is one. Here we are adding "com" but we could imagine others (such as "scala"), or even other prefixes "com.netflix".

That said, I think the real "solution" is to strong discourage or even scalafix away glob imports.

non:
@johnynek Yeah I'm thinking I will create a mapping of known roots, start it empty, and allow it to be populated by an environment variable (easiest way to plumb configuration through the tool, driver program, etc).

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/Entity.scala
non: @jklukas Alright, see what you think. I made it configurable.

jklukas (approved):
This looks great. I breathe easier having an interface for configuring these special values.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DriverApplication.scala
```diff
@@ -92,6 +92,22 @@ abstract class DriverApplication extends IOApp {
...
+  private[this] def isValidTld(s: String): Boolean =
+    s.matches("^[a-z]+$")
+
+  private[this] def setSpecialTlds(specialTldsEnv: String): IO[Unit] = {
+    val names = specialTldsEnv.split(',').toList
+    if (names.forall(isValidTld)) {
+      IO(Entity.setSpecialTlds(names))
+    } else {
+      val invalid = names.filter(s => !isValidTld(s))
+      IO.raiseError(new Exception(s"invalid TLDs: $invalid"))
```
jklukas, line 104: Test failures may be showing that we raise an error in the case of `specialTldsEnv` being empty?

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/Entity.scala
```diff
@@ -37,6 +37,48 @@ final case class Entity(parts: Vector[String]) {
...
+  //    import com.foo.com.bar.Qux
+  //    import com.acme.shadow.com.google.Dingus
+  //
+  // We could also handle other common TLDs such as "net" and "org"
+  // the same way but "com" is the most common and one of the least
+  // likely to occur as a "split import".
+  //
+  // This feature is disabled by default, and enabled in the driver
+  // application using the environment variable BZL_GEN_SPECIAL_TLDS.
+  private var specialTlds: Map[String, Entity.Resolved] =
```
johnynek, line 67: Can we use a mutex to protect this?

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/DriverApplication.scala
```diff
@@ -107,6 +123,8 @@ abstract class DriverApplication extends IOApp {
     ).mapN {
       (inputs, workingDirectory, label_or_repo_path, outputPath, sequential) =>
         for {
+          specialTlds <- getEnv("BZL_GEN_SPECIAL_TLDS")
+          _ <- setSpecialTlds(specialTlds.getOrElse(""))
```
johnynek, line 127: Why not: `specialTlds.traverse_(setSpecialTlds)`

Might also fix the test errors.

non: I tried this but actually I think it's also worth treating the empty string similarly, so I ultimately used a different approach.

language_generators/scala-defref-extractor/src/main/scala/io/bazeltools/buildgen/shared/Entity.scala
non: I ended up choosing a different design to minimize mutability.

non:
I ended up making the special TLDs a property of the extractor set at construction time.

This removes any concerns about mutability/locking, and also makes it a bit more natural.

I added tests to confirm that extractors behave differently with different sets of special TLDs.

README.md
```diff
@@ -184,6 +184,18 @@ In some situations, like for Protocol Buffer schemas, we want to generate second
 }
 ```
 
+### Heuristics
+
+Wildcard imports in Java and Scala can be expensive to resolve, since every subsequent import might be relative to the previous wildcard.
```
som-snytt, line 189: Java imports are always rooted. (This comes up when scalac reads Java sources.) Sorry I don't know the context, so my comment may be irrelevant.

Issue #33, opened 2023-04-14
Make collapsing opt-in


eed3si9n:
## steps

- Import bunch of code from other build tools with more chunky modules
- Regenerate

## problem

Currently it's too easy for the generated `BUILD.bazel` to turn into a big ball of:

```python
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_library")

filegroup(
    name = "shared_files",
    srcs = glob(include = ["**/*.scala"]),
    visibility = ["//visibility:public"],
)

scala_library(
    name = "shared",
    srcs = [
        ":shared_files",
        "//foo/shared/ante:ante_files",
        "//foo/shared/auctor:auctor_files",
        "//foo/shared/auctorgen:auctorgen_files",
        "//foo/shared/bibendum:bibendum_files",
        "//foo/shared/data:data_files",
        "//foo/shared/iaculis:iaculis_files",
        "//foo/shared/labelgen:labelgen_files",
        "//foo/shared/lectus/adaptor:adaptor_files",
        "//foo/shared/lectus:lectus_files",
        "//foo/shared/model:model_files",
        "//foo/shared/params:params_files",
        "//foo/shared/phasellus/placerat/lectus:lectus_files",
        "//foo/shared/phasellus/placerat:placerat_files",
        "//foo/shared/phasellus:phasellus_files",
        "//foo/shared/utils:utils_files",
    ],
    deps = [
       // huge list of 3rdparty dependencies
    ]
    ....
```

While it might be desirable at first to make one-big `scala_library` etc, the build is not healthy and hard to maintain and brittle, especially during migration, because transitively many targets would end up depending on the _huge list of 3rdparty dependencies_.

## expectation

We should pass in an allow-list of 1:1:1 rule violators (in this case all the subdirectories above), so at least we are aware when it happens, and we can code review any new introduction of bad targets.

weichunstevelee:
+1, this would be really helpful and help with build time. It's hard to instruct people to stick to 1:1:1 and also be aware of not having cricular deps between targets (which will then force the build generator to rollup and generate a big ball target). Would be nice to force in the tooling

ianoc:
[...]

PR #384, opened 2025-02-26
Make collapsing (kitchen sink target creation) opt-in

eed3si9n:
Fixes https://github.com/bazeltools/bzl-gen-build/issues/33

## Problem
People regularly create circular import at the package level partly because our tooling doesn't tell them that it's working around this bad behavior.

## Solution
This introduces a configuration called circular_dependency_allow_list, which opts in to collapsing for specific package prefixes,
which hopefully acts as a ratchet that we can prevent further bad behaviors.

## Note
Example error message:

```
Error: Circular dependency found in the package src/main/java/com/example/cycle.
  src/main/java/com/example, src/main/java/com/example/cycle
  Resolve the cycle, or opt in to collapsing this into a higher-level build target by adding circular_dependency_allow_list in the module config JSON.
  Note that creating such aggregate target across multiple directories will slow down the build anytime you make source change in the area.
```


crates/driver/src/build_graph.rs
```diff
@@ -486,14 +486,27 @@ impl GraphState {
...
                 if outbound_edge == node {
-                    return true;
+                    if let Some(node_label) = self.get_node_label(&node) {
+                        if !kitchen_sink_prefixes.iter().any(|p| node_label.starts_with(p)) {
+                            let mut cycle_nodes: Vec<_> = inner_visited.iter().cloned().collect();
+                            cycle_nodes.sort();
+                            
+                            return Err(anyhow::anyhow!("Circular dependency found in the package {}.
+  {}
+  Resolve the cycle, or allow the cycle this by adding kitchen_sink_prefixes in the module config JSON.",
```
jklukas, line 496:
```suggestion
  Resolve the cycle, or opt in to collapsing this into a higher-level build target by adding kitchen_sink_prefixes in the module config JSON.",
```

jklukas (approved):
We should provide some documentation for this feature, minimally by including it in the example module config in the README (https://github.com/bazeltools/bzl-gen-build?tab=readme-ov-file#module-config).

I'm not sure how obvious the "kitchen sink" name here will be to users, so consider renaming this option to something like `collapse_cycles`.

crates/driver/src/build_graph.rs
```diff
@@ -506,11 +523,16 @@ impl GraphState {
                 if no_loops.contains(node) {
                     continue;
                 }
-                if self.in_cycle(*node) {
-                    incompress = Some(*node);
-                    break 'outer_loop;
-                } else {
-                    no_loops.insert(*node);
+                match self.in_cycle(*node, circular_dependency_allow_list) {
```
johnynek, line 526: I think you can do:

```rust
let cycle = self.in_cycle(*node, circular_dependency_allow_list)?;
if cycle {

}
else ...
```
or:
```rust
match self.in_cycle(*node, circular_dependency_allow_list) {
  Ok(true) =>  {
                            incompress = Some(*node);
                            break 'outer_loop;
  }
  Ok(false) => no_loops.insert(*node);
  Err(e) => return Err(e)
}
```
I tend to think the `?` one is more idiomatic.

johnynek: actually, I think `if self.in_cycle(*node, circular_dependency_allow_list)? {` will work if you want to make a more minimal change, and it might just be a one liner at this point.

eed3si9n: Yea, `?` totally works fine - https://github.com/bazeltools/bzl-gen-build/pull/384/commits/a169ea0a26e328bd78843b078e6f0bc33cfc845a

PR #389, opened 2025-03-04
Document comparison to gazelle

jklukas:


johnynek:
This doesn't seem like a comparison. I honestly don't know the differences.

When @ianoc and I started this work the question was: can we even do a reasonable job of generating scala builds from source without compiling because many in the past have said that would never be practical.

Also it was informed by our experience that contributing to a Google open source project can be very difficult: they often have a strong prior of ignoring or rejecting PRs so they expect a lot of work on each PR to convince them of the value of the work or the style of your approach.

Lastly, for native code we preferred to use rust over go.

So the real difference is we could merge our own PRs (and didn't have to write go).

PR #426, opened 2025-09-05
Empty commit

jklukas:


jklukas:
cc @eed3si9n

eed3si9n:
GitHub maybe changed to M1/Aarch64/ARM64 macOS, and we might need to either downgrade the image, or upgrade some libraries.

jklukas:
@eed3si9n I pinned to `macos-14` and now all tests are passing. We'll need to change the "required" set of tests since "macos-latest" is in the name of 2 required tests (the ones currently showing in yellow as "Expected".

Once this is approved, I'll update the branch protection config for the new names.

